PR #46, opened 2018-02-14
Improvement: remove usage of !important to remove the padding of nodes during  search

Gregcop1:
feat: remove cascading `!important` in css for hidden nodes (it's repaced by JS conditional style  and allow to  override this behaviour easily)
feat: add a "keepTreeOnSearch" prop to display results as a tree instead of flattening them (parent that don't match but with children which match are displayed with a specific class (ATM, it reduces opacity too))

### DIFF
--- a/README.md
+++ b/README.md
@@ -215,6 +215,13 @@ Type: `string`
 
 The text to display as placeholder on the search box. Defaults to `Choose...`
 
+### keepTreeOnSearch
+
+
+Type: `bool`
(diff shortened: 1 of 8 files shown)


mrchief:
@Gregcop1 Thanks for sending this PR. `keepTreeOnSearch` is a nice addition. 

Just running few tests locally. If everything checks out, I'll merge this soon. 👍

PR #50, opened 2018-02-20
feat: disable selection on specific nodes

Gregcop1:


Gregcop1:
I had to revamp "tree-node" a bit to respect codeclimate rule of (25 lines per function)

mrchief:
Looks good. One question though: what happens to children if a parent is disabled?

Gregcop1:
ATM, it does nothing. You can toggle the parent and select/deselect children. IMHO, if we want to keep customization, user has to specify that children are disabled too.

mrchief:
I think we should keep it consistent with how `checked` works. If you disable parent, children are automatically disabled (disabling at "levels"?). And children nodes override the parent anyway (so both use cases can be supported). This is how selection/deselection works as well.

Gregcop1:
Ok, so if a node doesn't have "disabled" property, it inherits the property of its parent, but if it has "disabled" property, I take this value?
If so, I will work on it (maybe today)

mrchief:
Yes, the same idea essentially. But also remember that checking for node's immediate parent may or may not be sufficient. Take a look at [setInitialCheckState](https://github.com/dowjones/react-dropdown-tree-select/blob/master/src/tree-manager/index.js#L91). I think we can combine the checked/disabled lookup there (and avoid dual walk up the tree) and set properties appropriately. That'll fix the tests as well.

src/tree-manager/index.js
```diff
@@ -85,29 +85,37 @@ class TreeManager {
...
   /**
   * If the node didn't specify anything on its own
-  * figure out the initial state based on parent selections
+  * figure out the initial state for checked and disabled based on parent
   * @param {object} node [description]
   */
-  setInitialCheckState (node) {
-    if (node.checked === undefined) node.checked = this.getNodeCheckedState(node)
+  setInitialStatus (node) {
+    if (node.checked === undefined || node.disabled === undefined) {
```
mrchief: I think this check should not be done. If either `checked` or `disabled` is set, it'll skip assignment for both props. I guess just having:

```
const { parentCheckState, parentDisabledState } = this.getNodeStatus(node)
if (node.checked === undefined) node.checked = node.checked || parentCheckState
if (node.disabled === undefined) node.disabled = node.disabled || parentDisabledState
```

would suffice (and work correctly even if one is set and the other is not).

Gregcop1:
@mrchief I don't know how to fix codeclimate here, do you have any idea?

src/tree-manager/index.js
```diff
@@ -85,29 +85,37 @@ class TreeManager {
...
    * @param  {[type]} tree    [description]
    * @return {[type]}         [description]
    */
-  getNodeCheckedState (node) {
+  getNodeStatus (node) {
     let parentCheckState = false
+    let parentDisabledState = false
     let parent = node._parent
-    while (parent && !parentCheckState) {
+    while (parent && (!parentCheckState || !parentDisabledState)) {
```
mrchief: One prop may short-circuit the other. I don't think we can avoid multiple traversals (one per prop) here (with current traversal logic).

mrchief:
> I don't know how to fix codeclimate here, do you have any idea?

I think I do. I'll give it a try tonight if you don't mind.

src/tree-manager/index.js
mrchief: Don't worry about this part for now. I think fixing the Cognitive Complexity will eliminate this problem as well.

Gregcop1: Nope, it's ok. It continues until there's no more parent or both (check and disabled) haven't been setted to true. So var are not supposed to short-circuit each other.

mrchief:
@Gregcop1  Can you move your branch to this repo (I added you as a collaborator)? Either that or add me as collaborator on your fork. I cannot push commits to this PR otherwise.

Gregcop1:
I've just push it to this repo but I can't change source branch. Maybe you can as Admin.
BTW, thanks for the invitation ;)

mrchief:
I can't change the source either. What's more weird is that I cannot open a PR from this repo's branch as github already thinks #50 is open for this. To preserve accreditation, I would suggest you to close this PR, and then open a new one from this repo's branch. 

I've pushed my changes to the branch already. I feel we need more test cases though (like the [failing test](https://github.com/dowjones/react-dropdown-tree-select/commit/2351e36cd1699a32cbe00952919df9cbc2c9a7bb) which I think will fail in master as well). But take a look and let me know your thoughts.

Gregcop1:
You're right. I've just closed and reopened a new PR with your branch. Thanks for your help

PR #56, opened 2018-03-05
feat: embed children and leaves count on node and tag component to allow a better customisation via CSS

Gregcop1:
Allow customization like this with small piece of CSS:
capture d ecran de 2018-03-05 16-56-40

@mrchief Keep me informed if you want me to change implementation.

### DIFF
--- a/src/input/index.js
+++ b/src/input/index.js
@@ -9,10 +9,15 @@ const cx = cn.bind(styles)
 
 const getTags = (tags = [], onDelete) => {
   return tags.map((tag, i) => {
-    const {_id, label, tagClassName} = tag
+    const {_id, _children, _nbLeaves, label, tagClassName} = tag
     return (
       <li className={cx('tag-item', tagClassName)} key={`tag-${i}`}>
-        <Tag label={label} id={_id} onDelete={onDelete}/>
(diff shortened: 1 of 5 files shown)


mrchief:
I like the idea of expanding the control to provide additional info. However, I think it'd be far more useful to expose an interface that lets users surface arbitrary data and any amount of such data. And it'd be an added bonus if the control doesn't do any extra processing (that may or may not be used by everyone).

With that said, take a look at branch data-attributes. I tried to cook up something towards those points. 

In your case, you'd need to calculate leaf count by walking the tree before passing it to this control. This decouples the leaf count logic from the control. 

It also ensures everybody doesn't pay a performance penalty (counting leaves or any such computation) - only those who really need it incur the overhead. 

Thoughts?

mrchief:
@Gregcop1 Did you get a chance to look at this?

Gregcop1:
I'm sorry. I was off several days. I didn't watch out what you did yet but it's look promising.

mrchief:
No worries. We all have lives outside of github! Just let me know if fits your purpose whenever you get a chance. Feel free to make/propose changes.

PR #62, opened 2018-03-15
feat: Add no matches text in parameter

toofff:
I put this parametrable text so that it can be translated into any language.

Does this feature seem interesting to you for your plugin?


toofff:
@mrchief ok i'm looking at this today

PR #65, opened 2018-03-16
feat: Allow arbitrary data attributes to be set

### LINKED ISSUE #53: question: embed some node info on tag
The idea is to have children and leaf count on each node and to embed those infos on node and tag items via custom attributes ("data-nb-children", "data-nb-leaf" for ex.) so user can display them with a bit of css if he need it.

Displaying something like that 
capture d ecran de 2018-02-26 15-41-18
is done via `content: "("attr(data-nb-leaves)")";`

I don't know if you want something like that in your module but if so, I've got a PR for you ;)

mrchief:
Fixes #53, Fixes #56


mrchief:
@Gregcop1 If this looks good to you, we can ship this (need to add docs though).

PR #74, opened 2018-03-29
Create .codeclimate.yml

davehenton:


.codeclimate.yml
```diff
@@ -0,0 +1,50 @@
...
+- "features/"
+- "**/node_modules/"
+- "script/"
+- "**/spec/"
+- "**/test/"
+- "**/tests/"
+- "Tests/"
+- "**/vendor/"
+- "**/*_test.go"
+- "**/*.d.ts"
```
mrchief: I think I need these exclude patterns:

```
exclude_patterns:
  - "docs/"
  - "snapshots/"
  - "**/node_modules/"
  - "**/*.test.js"
```

davehenton:
Hey @mrchief, quick update here. After looking at the docs for the eslint-plugin-react plugin [here](https://github.com/yannickcr/eslint-plugin-react#configuration), it seems that you need to explicitly enable JSX support in your ESLint config in order to define the rules properly. I've updated the `eslintrc.json` to account for that, and I've also updated the exclude patterns to reflect what you mentioned. Let me know if this works for you!

mrchief:
@davehenton Thanks for this update. airbnb config is supposed to enable JSX support, hence the omission. But having it here doesn't harm so 🤷‍♂️ 

Also, I had to disable the object curly rule to get around CC errors. When you merged master, you got this rule merged as well: `"object-curly-newline": 0`. 

Do you mind reverting this rule and pushing again (to ensure that the issue has indeed been fixed)?

PR #76, opened 2018-04-03
feat: Add possibility to clear search on node selection

Gregcop1:


mrchief:
Looks good. We should add this options to Options story under docs as well.

Gregcop1:
So how do you want to proceed. We merge this then I do an exemple in codesandbox and we merge again?

mrchief:
I use the options story as an end user test (this way we don’t leave anything to fate). 

Add an option there, test out if the component does what it’s supposed to and if everything checks out, I know I’m good to go. 

Bonus, we get free documentation with real live examples. 

Codesandbox is good but I don’t think we need to create one for every little flag. People can play with the base sandbox or can checkout docs for examples. 

Makes sense?

Oh wait, is the options story part of the partial state PR? If so, I guess I’ll have to merge it soon. I was hoping codeclimate support can fix the issues first but maybe I’ll skip it for that PR (`yarn lint` checks out so I know its GTG). I’m on a phone so I’ll let you know more tonight.

Gregcop1:
I don't see the options entry in my doc so I think it's in your current PR. (I was asking myself how you do local test with this process but now I know. :))
I review your PR right know and will update mine after your merge

Gregcop1:
@mrchief I've updated the "story" part

src/index.js
```diff
@@ -130,6 +131,9 @@ class DropdownTreeSelect extends Component {
     })
     if (this.props.simpleSelect) this.resetSearch()
     this.notifyChange(this.treeManager.getNodeById(id), tags)
+    if (this.props.clearSearchOnChange) {
```
mrchief: This can probably be combined with `if (this.props.simpleSelect)`

mrchief:
Would you mind running `yarn build:docs` and checking in the updated `docs/bundle.js`? Also note a minor comment.

Issue #94, opened 2018-04-20
Feature request: Parent select shouldn't select all children


will-izard:
Hi. I'm looking for a widget where a branch or leaf in the tree can be selected. Selecting a branch/parent does not mean the children are selected. Otherwise, this widget looks great! Is there any way you could add that as a configuration option? Or maybe I overlooked a way to do this already?
The use case is that you're selecting a category in a tree. You can select 1 or more categories, but the categories can be parent or child (ie branch or leaf) categories. 
Thanks!

mrchief:
If I'm understanding you correctly, I think this component already does what you're looking for in its simplest form.

If you don't set `showPartiallySelected` to true (it's false by default) then selecting a parent will give you only the parent node in `selectedNodes` param (visually it'll check all children to indicate the parent has been checked). If you select all children manually, then you get all children nodes in `selectedNodes` and not the parent. 

Is this what you're looking for?

will-izard:
Hi, thanks. It's this bit that I'm not keen on "visually it'll check all children to indicate the parent has been checked". That indicates that all the children will be stored as selected as well, which they are not.

will-izard:
Suggest simply show the parent as selected, not the children.

mrchief:
I think that'd be weird UX wise, at least with checkboxes still visible. Maybe if we didn't have checkboxes, it'd look OK... 

@toofff @Gregcop1 Thoughts?

will-izard:
I think you have a different use case in mind to me. Both behaviours are valid. My use case is that user wants to select a page on a website on which to show a given item/advert etc. The pages are the items in the tree. They can select a parent page - that doesn't mean that the item/advert will be show on the child pages - just on the parent page.
Hence, suggest this as a config option please.

mrchief:
I get the config part, and I agree with you that it should be configurable. What I'm asking is , do you still want to see checkboxes in that case?

will-izard:
Yes, for sure, I still need checkboxes. Why would I not? Are you refering to the single select case? That could be useful - if i check on checkbox, it unchecks any other selected checkbox. Is that what you mean?

r20:
I'm also interested in being able to select a parent without the "visual" selection of children.

will-izard:
I want the checkboxes still, as it can still be multi select. Would be good if the widget worked with a config flag to be single select and multiselect.
[...]

PR #103, opened 2018-05-14
fix: Parent check when selecting all children doesn't work

Gregcop1:
@mrchief I think there's a small regression since partial check. We walk through parents only if `showPartialState` is selected but this option should only affect the partial check but not parent selection.
ATM, if `showPartialState` is not used, when I select all children, the parent doesn't check itself. This PR fixes this.


mrchief:
Not a regression. This was done intentionally. This allows you to have a way to explicitly select individual child values (in case you need to distinguish between the use cases when you need parent or all children under parent _at_ that point in time).

`showPartialState` behaves differently because it needs to do so in order to have visual consistency for partial states.

Gregcop1:
OK. My bad. Do you want me to add an option for this or should we close this PR?

mrchief:
Let's close it for now. Don't want too many options for trivial effects. I'm pondering over doing something for #94. We can revisit this later.

Issue #108, opened 2018-05-24
Get all selected values (including childrens)


koko-ng:
Hi,
Is it possible to get all children values when a parent is selected? Or maybe to get the whole flattened tree and then filter it?
Btw, is it normal that `children` is undefined in `currentNode` or  `SelectedNodes` in the `onChange` callback?

mrchief:
> Get all selected values

The 2nd argument returns this info: `onChange(currentNode, selectedNodes)`

> Is it possible to get all children values when a parent is selected?

It's possible but it needs a little bit of upkeep on your side. 

 - Assign node paths to your tree via an [HOC](https://dowjones.github.io/react-dropdown-tree-select/#/tree-node-paths-hoc)
 - Take the node path (any extra props on the node object are returned as is) to grab the children from your input data.

> Or maybe to get the whole flattened tree and then filter it?

The flattened tree is an implementation detail and not part of the public API. You should not depend on it. Also, there is nothing that exposes the flat tree.

> Btw, is it normal that children is undefined in currentNode or SelectedNodes in the onChange callback?

Yes. `children` prop comes from your original input data. Deleting the prop sacrifices certain optimizations by v8, hence it's marked as undefined since the component doesn't need it after initialization (this keeps the memory requirements light).

Tip: You can look at `_children` prop which contains ids for child nodes. They always follow this pattern: `level-child#` (which repeats as you go deeper). Underscore props are somewhat private, as I'm not sure if they should be part of the public API. So use it with caution.

koko-ng:
Thanks for you answer, I tried your solution with the HOC. It works like a charm :). I put a list of all its childrens on each elements, maybe not the most efficient way of doing this, but it works.

mrchief:
> I put a list of all its childrens on each elements, maybe not the most efficient way of doing this, but it works.


I knew letting misc. props bubble up will come handy some day!  :innocent:

FredyC:
Sorry for reviving the older issue, but I don't get it. If `selectedNodes` contains only top-level nodes and `_children` are non-leaf only, I am confused how can I get those selected leafs? Even if I store node paths there, I still don't know which leafs are selected. What am I missing, please?

mrchief:
> If selectedNodes contains only top-level nodes and _children are non-leaf only, I am confused how can I get those selected leafs? 

If you select a top-level node, it implies all children are selected. Does this help clear the confusion?

FredyC:
[...]

PR #109, opened 2018-05-24
feat: Add the whole tree in onChange callback

koko-ng:
See #108

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -148,7 +148,7 @@ class DropdownTreeSelect extends Component {
     }
 
     this.setState(nextState)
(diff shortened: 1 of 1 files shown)


mrchief:
Thanks for taking the time to create this PR.

This is not something I'd agree to - simply because the internal tree is an implementation detail. Exposing it makes it part of the public API which imposes restrictions on what I can and cannot do with it. 

Also, I'd like to keep changes backward compatible for this component (not saying there will never be any breaking changes) for most part, so exposing something like that is pretty constraining towards that goal.

koko-ng:
Yes, I totally got your point, the solution you suggested is *way* better. (#108)

PR #120, opened 2018-06-20
Update webpack config to use babel presets and plugins

mattbague:
Main reason for this change is to transpile ES6 code. I noticed you guys had babel presets/plugins defined in the `package.json` so I was hoping they were intended to be used.

Note that this does make the final file generated bigger (40K to 112K). This can be reduced to 48K if you don't need the following babel plugins applied:
* istanbul
* transform-class-properties
* transform-react-remove-prop-types

This change would help fix an issue with a project we're working on. Our current test infrastructure doesn't like ES6 code sadly :(

Thanks!

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -39,7 +39,17 @@ module.exports = {
     rules: [
       {
         test: /\.js$/,
-        loaders: ['babel-loader'],
+        use: {
+          loader: 'babel-loader',
+          options: {
+            presets: ['es2015', 'react', 'stage-0'],
+            plugins: [
+              'istanbul',
+              'transform-class-properties',
+              'transform-react-remove-prop-types'
+            ]
+          }
+        },
         include: path.join(__dirname, 'src'),
         exclude: /node_modules/
       },


mrchief:
Thanks for reporting this. 

The settings defined in package.json are for `ava` and do not participate in webpack compilation. webpack compilation reads the settings from `.babelrc` which is suppose to transpile to ES5 codebase. 

The existing babelrc has similar options to your changes except for `["es2015", { "modules": false }]` (and ignoring prop-types removal).

Would you be able to share some specific errors or details as to what exactly is failing and how? 

I've been planning to support ES6 (via `jsnext:main`) and ES5 by creating separate builds. So maybe what you share will help there as well.

mattbague:
Ahh I see. What we were seeing was when we got js file from this URL: https://unpkg.com/react-dropdown-tree-select@1.12.0/dist/react-dropdown-tree-select.js

It had this line: `const n=e.filter(t);` So the use of `const` was causing us some issues since that was introduced in ES6.

mrchief:
`const` is supported in all current browsers. Are you using IE10 or below or similar browser?

mrchief:
Hmm. To keep code bloat to a minimum, I try to support only current browsers. The separate builds will still not help in your case unless I target lower browsers. 

I wonder how your changes fixes this issue though. es2015 preset doesn't transform `const` to `var` unless you add more plugins. 

What browser are you using?

mattbague:
Looked into it a bit more and figured out that it's the [classnames](https://www.npmjs.com/package/classnames) dependency that's causing the problem. I'm just gonna close this pull request out.

PR #139, opened 2018-07-26
Fix partial select and outside click

### LINKED ISSUE #137: Incorrect behavior while using showPartiallySelected property
Hi,

Thanks for this awesome component 😄 . I am using it to show multiple hierarchies on a page and the data for these hierarchies is pulled from a web API. When the data is loaded then the component is rendered. 

The problem which I am facing is when I use `showPartiallySelected` property then the root node(s) become automatically checked in the UI and I see the pill(s) for all of them. I have explicitly marked the property `checked` as `false` in the data but that doesn't help also. 

Could you please look into this issue? Let me know if you need more details.

PS: I cloned the repo and used `showPartiallySelected` in the `Large Tree` example and same thing happens. The root node is checked by default. However, it works fine in the `Default Values` example. 🤔 

Thanks,
Priyanshu

priyanshu92:
Closes #137


src/index.js
```diff
@@ -100,7 +100,7 @@ class DropdownTreeSelect extends Component {
   }
 
   handleOutsideClick = e => {
-    if (!isOutsideClick(e)) {
+    if (!isOutsideClick(e, this.props.className)) {
```
mrchief: What does `this.props.className` do here?

src/utils/isOutsideClick.js
```diff
@@ -12,7 +12,8 @@ const getPath = e => {
   return path
 }
 
-export default e => {
+export default (e, className) => {
```
mrchief: Not sure what this has to do with the empty children check. I've a feeling this is either test stuff or you're trying to solve a different issue - in which case it should be separate discussion. Can you elaborate on this please?

src/tree-manager/flatten-tree.js
```diff
@@ -159,7 +161,7 @@ function walkNodes({ nodes, list = new Map(), parent, depth = 0, simple, showPar
         node.partial = getPartialState(node)
 
         // re-check if all children are checked. if so, check thyself
-        if (node.children.every(c => c.checked)) {
+        if (!isEmpty(node.children) && node.children.every(c => c.checked)) {
```
mrchief: I would also add a unit test for this.

src/utils/isOutsideClick.js
priyanshu92: Yeah this is a separate issue. Do you want me to create another issue for it?

priyanshu92: When we have more than one dropdowns on the page then if we open one dropdown and then open another, the first one doesn't close because we are checking the `outsideClick` based on CSS class and since both have same classes the dropdown doesn't close on opening another one. And if you click outside then all the dropdown closes at once.


If you want, I can create a separate issue and PR for this.

mrchief: Sounds good! Will merge it soon once I verify locally.

mrchief: While testing this particular change, I found that this works only if the user assigns a unique classname to each instance. Otherwise, it's the same behavior.

priyanshu92: Right. So it keeps the existing behavior and also gives the user a way to modify it.

mrchief: That wasn't the point I was trying to make but I like the way you put it! :)

mrchief:
Can you resolve the conflicts and push again? Easiest would be to take either base or target's version of bundle.js to resolve conflict, and then run `yarn build:docs`, to update the bundle.js to it's most recent state.

mrchief:
Crap, the commit message doesn't follow semantic conventions. Will have to publish manually! :)

priyanshu92:
How should it be written?

mrchief:
Same as you write the commit messages. But unlike those commit messages (which are linted by git hooks), GitHub's PR have no such protection. 

And their [docs](https://github.com/semantic-release/semantic-release/#commit-message-format) are bit misleading. They care about the PR title more than the commit messages.

mrchief:
I'll get this released over the weekend - I need to sort out some other issues with the whole automated release process as well.

mrchief:
v1.12.4 is live now!

PR #141, opened 2018-08-01
fix: Do not toggle dropwdown on pill removal

### LINKED ISSUE #134: Toggle dropdown visibility on click remove tag
In this example - https://dowjones.github.io/react-dropdown-tree-select/#/story/basic-no-extra-styles
If you clicked on X tag, it's toggle dropdown visibility. But in other example all works correctly.

Thx!

priyanshu92:
Fixes #134 

Prevents the dropdown from toggling again and again when the selected tags are removed.


src/tag/index.js
```diff
@@ -25,7 +25,7 @@ class Tag extends PureComponent {
     return (
       <span className={cx('tag')}>
         {label}
-        <button onClick={this.handleClick} className={cx('tag-remove')} type="button">
+        <button onClick={e => this.handleClick(e)} className={cx('tag-remove')} type="button">
```
mrchief: This can be simply `onClick={this.handleClick}` (avoids creating a function on every render). I wonder why linting didn't flag this.

src/tag/index.js
```diff
@@ -13,9 +13,9 @@ class Tag extends PureComponent {
     onDelete: PropTypes.func
   }
 
-  handleClick = () => {
+  handleClick = e => {
     const { id, onDelete } = this.props
-
+    e.stopPropagation()
```
mrchief: Don't remember why but I think we also need to add `e.stopImmediatePropagation()` too.

src/tag/index.js
priyanshu92: Okay. I'll update it.

src/tag/index.js
priyanshu92: Right. We can use `stopImmediatePropagation()` because it implicitly calls `stopPropagation()`. I'll update the code.

Issue #148, opened 2018-08-09
Bug: Tree collapses when element selected beyond first InfiniteScroll load


nikolap:
Behaviour:
Clicking on any tree node after the first InfiniteScroll load (after you pass element 100) closes the tree. The item is correctly selected/deselected, however.

Intended behaviour:
Any element selection click (unless simple select true) should not hide the dropdown (i.e. tree). In this case, scrolling past the first InfiniteScroll load and clicking on any of the tree nodes should keep the tree and scroll position as it is.

To reproduce:
You can use the large-tree, for example (https://dowjones.github.io/react-dropdown-tree-select/#/story/large-tree)

1. Search for '1'
2. Scroll until the next batch of tree nodes loads
3. Click on any node
4. The node will be selected **but** the tree will be closed

Tested on Firefox 62.0b15 

Amazing job with this library, finding it very useful. Let me know if there's any clarification or debugging I can help with.

nikolap:
Strange, Test 1 should have caused the bug:

![peek 2018-08-10 07-04](https://user-images.githubusercontent.com/4504265/43955063-a6f32b1a-9c6c-11e8-8015-f00cf1037f36.gif)

Anyways, I did some more digging and had it reproduced with the same version of Firefox (see gif above). So figured it must be a browser or addon thing and tried Chromium which worked just fine! So I went back and tried an old version of Firefox (v59), fresh and clean install. It had the same issue.

Long story short -- appears to be an issue on Firefox.

mrchief:
Seems like something’s different in Firefox in regards to event handling. Inside click should not close the dropdown.

nikolap:
Strangely, it doesn't close the dropdown *until* you scroll past the initial InfiniteScroll break.

![peek 2018-08-10 07-40](https://user-images.githubusercontent.com/4504265/43956078-bfe508d8-9c70-11e8-9e80-3efb36ff5a6f.gif)

nikolap:
The closing/opening issue appeared to be due to the fallback path resolution in `getPath` (isOutsideClick.js). To simplify matters, Firefox has a composedPath function available and works as a good fallback before going to manually composing the path.

Note 1: I did not investigate why/where the fallback path resolution was failing in FF.
Note 2: There is a secondary bug (also browser specific) wherein the scroll position is reset incorrectly on Firefox, though correctly on Chromium.

PR #154, opened 2018-08-14
fix: Firefox outsideClick verification fails on certain elements (#148)

nikolap:
#148 partial fix. Now the tree display remains open, however scroll position is incorrectly reset

### DIFF
--- a/src/utils/isOutsideClick.js
+++ b/src/utils/isOutsideClick.js
@@ -1,5 +1,5 @@
 const getPath = e => {
-  let easyPath = e.path || (e.composedPath && e.composedPath());
+  const easyPath = e.path || (e.composedPath && e.composedPath())
   if (easyPath) return easyPath
 
   let elem = e.target
--- a/src/utils/isOutsideClick.js
+++ b/src/utils/isOutsideClick.js
@@ -1,5 +1,6 @@
 const getPath = e => {
-  if (e.path) return e.path
+  let easyPath = e.path || (e.composedPath && e.composedPath());
+  if (easyPath) return easyPath
 
   let elem = e.target
   const path = [elem]


nikolap:
Sure, send them along, more than happy to lend a hand if I can :)

mrchief:
Ok so there are 2 ideas I wanted to explore. But before I lay out implementation details, here are the some of objectives I wanted to fulfill:

 - The outside click should be able to distinguish between multiple instances of the component on the same page. Basically improvise [this situation](https://github.com/dowjones/react-dropdown-tree-select/pull/139#discussion_r207079030) even when no unique classnames are passed.

 - The outside click should work on all browsers (including the less beloved IE). I'm OK with IE v11 and above if it comes to that.

 - If possible, do not involve any 3rd party libs (unless it's very very lean and introduces no significant increase in bundle size)

So with those golas in mind, what I wanted to try is:

 - [descendant check](https://github.com/dowjones/react-dropdown-tree-select/blob/bbd5c2ca62cb4f563dc49c8a7bcf680fa872370f/src/index.js#L89). We had this before but it wasn't working reliably. Don't remember why or how.  I think using `capture: true` on the document event listener might help. I'd also like to deregister the listeners on unmount and maybe play with onMouseDown/onMouseUp instead of onClick.

 - [invisible overlay](https://vncz.js.org/outsideclick/): this a neat trick that involves displaying a full screen modal with `z-index: -1` and listening for clicks on that div. The idea seems interesting and avoids global event handlers; but dealing with full screen modals is always tricky, especially with multiple components. Not to mention the caller app will most likely have its own modal overlays. 

I guess having an `OutsideClickHandler` that just takes a child component and deals with all the outside click handling might be better (rather than having it as part of `index.js` like I had above). 

So there you go. These are some of the ideas I had going. You're more than welcome to bring your own into the mix!

nikolap:
Haven't played around with either approach yet, but another option I've seen is the way semantic UI react checks for clicks via appending a data-attribute to a ref:

https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/lib/doesNodeContainClick.js
https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/modules/Dropdown/Dropdown.js#L635

Let me know your thoughts...

mrchief:
It's the same concept as descendant check but I'm not sure why it's so complex. Maybe they need to handle some use cases that require such logic.

But appending data-attribute seems like a moot op to me because we already have a ref node. Further, looking up by data-attribute will return all instances (basically same problem as non unique classname).

nikolap:
Sorry, haven't had any chance to look into this recently. Feel free to go on without me. Otherwise, will have some time the following week or two

nikolap:
Briefly looked at the previous approach via descendant check/Node.contains. The issue appears to be that the ref does not contain the updated dom information following the next call for InfiniteScroll. 

I think this may be tied to how the InfiniteScroll component manages state. Or perhaps how it's implemented here. If we resolve that it should hopefully a) let us use the Node.contains as  the check approach for outside click and b) prevent the scroll reset on FF and the scroll re-load on Chrome.

mrchief:
Found a workaround for the time being: https://github.com/dowjones/react-dropdown-tree-select/commit/2a1302b9113805372131a0fd2386ed1831c87ae2. I tested on firefox and seems to be working. While this issue is definitely worth pursuing, we can release this fix to get you unblocked in the interim.

nikolap:
Nice one! Thank you! I'll leave it to you to decide whether to release or not -- I can live with it for now while we look for a more permanent fix.

mrchief:
Fixed via #177

PR #162, opened 2018-08-24
feat: Allow passing function to override node render

MomchilGorchev:
This is a feature proposal of allowing functions to be passed to override the node and toggle icon renders. This provides easy way of customisation. Please review, test and let me know if im not missing something, will update the docs if it gets green light 👍


mrchief:
@MomchilGorchev Thanks for this PR. 

This is an advanced feature and we'd need a few things if we want to support this properly:

 - Documentation - What these do, how to write one of these, what are the rules, when to use one and so on.
 - Live Examples
 - Tests - Node and Icon renderer are integral to proper working of the component, so it needs to have proper tests.

Let me know if you have any questions.

Issue #187, opened 2018-10-16
Can not disable input


akovacspentalog:
pass the disable prop from search input, to be able to disable the input when necessary.

mrchief:
@akovacspentalog Thanks for reporting this. This is a duplicate of #127. 

I have given it some thought and supporting just `disabled`  is straightforward. Unfortunately, I haven't had the time to put in this yet. 

In my opinion, I think `disabled` could do one of the following:

 - disables search input, as well as dropdown (no dropdown on click)
 - disables search input, disables checkboxes in the dropdown but allows opening/closing of dropdown

I'm leaning towards the first since that resembles how the native `select` and `input` behave. The later could be how `readonly` works.

If you (or anyone looking to contribute) would like to send a PR for this, I'd very much appreciate that.

itrombitas:
Hi there,

I work with @akovacspentalog and we're looking to use this component in our project, but we'd also need to be able to disable it.

I've got a solution ready to go for option 1: 

> - disables search input, as well as dropdown (no dropdown on click)

Could you grant me access to push a branch and create a new PR with my changes?

mrchief:
@itrombitas can you fork the repo and send me PR? If you haven’t done this before, you can follow this guide 
https://guides.github.com/activities/forking/. 

Thanks for your contribution.

PR #188, opened 2018-10-17
feat: disable input

itrombitas:
Solution for [187](https://github.com/dowjones/react-dropdown-tree-select/issues/187)

I've had to update package.json to get the specific versions ava and babel, otherwise I couldn't get the tests to run properly. I've actually just checked this from [PR186](https://github.com/dowjones/react-dropdown-tree-select/pull/186)

disable search input and dropdown (no dropdown on click)
tests to cover new features
update storybook options

If there's anything i missed or need to do, please let me know.


src/index.js
```diff
@@ -175,7 +179,7 @@ class DropdownTreeSelect extends Component {
   render() {
     const dropdownTriggerClassname = cx({
       'dropdown-trigger': true,
-      arrow: true,
+      arrow: !this.props.disabled,
```
mrchief: I think we should leave the arrow as is but add another class `disabled` for `disabled` case. This accomplishes few things:

 - keeps the current behavior as is
 - the arrow should not disappear, rather appear as disabled for a better UX
 - allows users to extend or style differently when the control is disabled (since it's not a native control, they cannot use CSS attribute selector)

Also, I think we need some disabled styling for arrow like greying it out and setting cursor to not-allowed.

src/input/index.js
```diff
@@ -47,6 +48,7 @@ class Input extends PureComponent {
         <li className={cx('tag-item')}>
           <input
             type="text"
+            disabled={this.props.disabled}
```
mrchief: Can you move it to the props destructuring statement at the beginning - to keep things consistent?

```
const { ..., disabled, ...} = this.props
...
disabled={disabled}
...
```

src/index.js
```diff
@@ -84,6 +85,9 @@ class DropdownTreeSelect extends Component {
 
   handleClick = () => {
     this.setState(prevState => {
+      if (this.props.disabled) {
+        return { showDropdown: false }
+      }
```
mrchief: I think we can save some unnecessary render cycles if we simply add a no-op click handler for disabled

```
<a className={dropdownTriggerClassname} onClick={!this.props.disabled && this.handleClick}>
```

mrchief:
Awesome. Changes look good on paper. I'll run this branch locally and if everything checks out, will merge it. Stay tuned!

Issue #212, opened 2019-02-27
Different behavior (selecting the checkbox VS clicking the label) with multiple instances on the same page


stoberov:
**Describe the bug**

I'm currently using version 1.16.0 and have a page with 2 DropdownTreeSelect instances on it. 
Dropdown 1 works as expected. Dropdown 2, however, works as expected only when selecting the **checkboxes** - if you clicked any of the Dropdown 2 **labels**, the change affects items in Dropdown 1 instead.

I've noticed that the behavior is not present with older version of the module - 1.10.0 and below. So I'm guessing it is introduced recently, or there is some extra prop that needs to be passed these days.

**To Reproduce**

Sample: https://codesandbox.io/embed/qlzpr7lzmj?fontsize=14

If you select A, B or C (either checkbox or label) - all good.
If you select D, E or F by checkbox - all good. 
**If you select D, E or F by clicking on the label (label = the actual letter visible on the screen, to the right of the checkbox), this will select A, B or C instead.**

**Expected behavior**
Clicking on D / E / F labels should select D / E / F. It should not select  A / B / C.

mrchief:
I think this [`id`](https://github.com/dowjones/react-dropdown-tree-select/blob/develop/src/tree-node/node-label.js#L52) may not be unique across instances.

mrchief:
I think we can simply get rid of the `htmlFor` attribute as the input is nested. It may throw some accessibility warning but it's not needed. That change was introduced later and that'd explain why older versions work properly.

mrchief:
See https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/label-has-for.md

> To fully cover 100% of assistive devices, you're encouraged to validate for both nesting and id.

But like I said, we can either remove it or generate it using some combination that is both unique and deterministic. `node.id` may be a good candidate and if not, I think removing it for now will also work.

stoberov:
> I think we can simply get rid of the `htmlFor` attribute as the input is nested. It may throw some accessibility warning but it's not needed. That change was introduced later and that'd explain why older versions work properly.

@mrchief, thanks for the speedy reply!
Indeed, I can confirm that removing `htmlFor` (did it manually in my node_modules folder) seem to fix the issue.
It would be wonderful if either this or that other idea for `node.id` flows into a new release.

mrchief:
> It would be wonderful if either this or that other idea for `node.id` flows into a new release.

[...]

PR #213, opened 2019-02-27
fix: Ensure node labels toggle inputs from their own instance

stoberov:
## What does it do?

- Comments out a CSS import of a file that hasn't been committed to the repo
- Removes htmlFor as described in #212


docs/src/stories/DefaultValues/index.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react'
 
 import DropdownTreeSelect from '../../../../src'
-import '../../../../dist/styles.css'
+// import '../../../../dist/styles.css'
```
mrchief: Did you run `yarn build`? This file gets created by the build and this import tells Webpack to bundle it for the docs.

mrchief:
Few things before we can get this merged:

 - Need to run tests and update snapshots (you can run `yarn test --update-snapshots`) and commit those files.
 - Need to run `yarn build:docs` to update the bundle for documentation site.
 - Finally, the PR title needs to follow conventional commit message specs.

docs/src/stories/DefaultValues/index.js
stoberov: Ow! 
Here's the thing - I cloned the repo, installed the modules and then (as per the README.md) immediately did a `npm run demo`. For some reason, that styles.css was not generated, hence console would throw an invalid import error. 
It makes full sense now & the css file gets created as expected. I've thus reverted the commit which commented out the line.
Thanks.

mrchief: `npm run demo` is for local development. It doesn't use the filesystem, rather uses the in-memory artifacts.

__snapshots__/src/tree-node/actions.test.js.md
```diff
@@ -8,22 +8,20 @@ Generated by [AVA](https://ava.li).
...
 > Snapshot 1
 
-    [
-      <Action
-        actionData={
-          {
-            action: 'action-0',
-            id: undefined,
-          }
+    <Action
```
mrchief: This file should not have changed. I'm not sure what happened. Are you using React 15 by any chance?

stoberov: Hey! 
Wow, you're super active, kudos :) 
I myself have been bumping my head against this for the past 30mins. Travis fails due to this test, and yeah - I haven't made anything odd to change the file, other than `yarn test --update-snapshots`.
React is at version @**16.8.3**

src/tree-node/node-label.js
```diff
@@ -49,7 +49,7 @@ class NodeLabel extends PureComponent {
     }
 
     return (
-      <label title={title || label} htmlFor={id}>
+      <label title={title || label}>
         <Checkbox
           name={id}
```
mrchief: Now I'm thinking, the `name={id}` is not unique in the page when you have multiple instances. This is bad HTML semantically. Had these been radio buttons, the problem would have been more severe. 

I added https://github.com/dowjones/react-dropdown-tree-select/issues/215 to keep track of it.

mrchief:
This is looking good. If we can figure out the issue with actions snapshot getting updated, I think it should be good to merge.

__snapshots__/src/tree-node/actions.test.js.md
mrchief: You're using a recent version which also is suspicious 😈 . I'm wondering if the new React is discarding array if it only has one element and treating it as a single dom node (I'm thinking for efficiency). 

To test the theory out, we could add one more item to the array in the [test](https://github.com/dowjones/react-dropdown-tree-select/blob/develop/src/tree-node/actions.test.js#L9) and see if it renders the array back in the snapshot.

stoberov: Hahaha :) It's never boring with JavaScript's dependencies! :) 
Indeed, I included one more dummy object in the actions array, rebuilt the snapshots and Travis does not complain any more :)

mrchief: The plot thickens. 🤔 

I wonder if it's React or ava. If it's React, then it means more issues for me down the line. 😕

stoberov:
Hey, super sorry to be a pest! Just checking if there is any chance for a release that maybe includes this PR some time next week? Thanks for a great module!

mrchief:
I was looking into the actions snapshot issue a bit. I tried upgrading to react 16.8.3 and re-ran the tests but found no issues. The snapshot still was the same. See the screenshot below

image

So I'm not sure why you got an updated snapshot. Can you try the following steps:

 - `rm -rf ./node_modules`
 - `yarn install`
 - `yarn test --update-snapshots`
 - commit updated snapshots and push

If all goes as expected, we should have only one updated snapshot ***after*** you push.

Travs build will not fail when you push updated snapshots.

ellinge:
The issue with snapshots is solved in #220 
Bumped enzyme*-versions. Really should use a lock-file to avoid these kind of issues.

Also a bug in with webpack-dev-middleware which needs you to use 3.6.0 if you don't want to get an error when running demo.

mrchief:
> Bumped enzyme*-versions. Really should use a lock-file to avoid these kinds of issues.

We do have `yarn.lock` but I think we need to pin down exact versions instead of caret ranges.

Anyway, thanks for looking into this!

Let's add @ellinge for maintenance /cc @allcontributors[bot]

mrchief:
@stoberov Do you want to go ahead with this or wait for #220 which would take care of the original issue?

stoberov:
> @stoberov Do you want to go ahead with this or wait for #220 which would take care of the original issue?

Hey! Think #220 is the better solution.

mrchief:
Thanks @stoberov! 

Closing in favor of #220

PR #217, opened 2019-03-12
feat: Added support for single select in tree dropdown

### LINKED ISSUE #119: Simple select with tree view structure
Hi Team,

I have a need to show the dropdown option in a hierarchy form and allow only to select one of them.
I tried passing keepTreeOnSearch and simpleSelect to the component but it is not working.
Do you know if there is a way to acheive this?

ellinge:
## What does it do?

Adds support for single select in the tree dropdown. Simple select ignores any children so this is a hybrid between the two.

Also ignores clicks for simple select when labels are disabled (only checked readonly)

The middle dropdown is radioSelect and the last one a simpleSelect
https://ellinge.github.io/react-dropdown-tree-select-test/#DevelopTemp-checkeddefault

Fixes #119 

## Type of change
- [x] Bug fix
- [x] New feature


src/index.js
```diff
@@ -71,8 +72,8 @@ class DropdownTreeSelect extends Component {
   }
 
   componentWillMount() {
-    const { data, simpleSelect, showPartiallySelected, hierarchical } = this.props
-    const tree = this.createList({ data, simpleSelect, showPartiallySelected, hierarchical })
+    const { data, simpleSelect, singleSelect, showPartiallySelected, hierarchical } = this.props
+    const tree = this.createList({ data, simpleSelect, singleSelect, showPartiallySelected, hierarchical })
```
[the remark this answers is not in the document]
ellinge: No a new issue

src/index.js
```diff
@@ -82,8 +83,8 @@ class DropdownTreeSelect extends Component {
   }
 
   componentWillReceiveProps(nextProps) {
-    const { data, simpleSelect, showPartiallySelected, hierarchical } = nextProps
-    const tree = this.createList({ data, simpleSelect, showPartiallySelected, hierarchical })
+    const { data, simpleSelect, singleSelect, showPartiallySelected, hierarchical } = nextProps
+    const tree = this.createList({ data, simpleSelect, singleSelect, showPartiallySelected, hierarchical })
```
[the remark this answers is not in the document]
ellinge: Not a new issue

ellinge:
@mrchief this one should be up for grabs. I did not change the behaviour for simpleSelect so if that is something you think we should do instead, it would require additional changes.

I also noticed on the storypage that simpleSelect is not working great with default values, which also applies to this solution. Only one default value should be possible and should be deselected one selecting something else.

mrchief:
> I did not change the behaviour for simpleSelect so if that is something you think we should do instead, it would require additional changes.

Just one question - with this change, can we manage `simpleSelect` without the need of an additional flag?

> I also noticed on the storypage that simpleSelect is not working great with default values, which also applies to this solution. Only one default value should be possible and should be deselected when selecting something else

Great! Thanks for taking notice and fixing it!

src/index.js
```diff
@@ -71,21 +78,16 @@ class DropdownTreeSelect extends Component {
...
   componentWillUnmount() {
     document.removeEventListener('click', this.handleOutsideClick, false)
   }
 
   componentWillReceiveProps(nextProps) {
-    const { data, simpleSelect, showPartiallySelected, hierarchical } = nextProps
-    const tree = this.createList({ data, simpleSelect, showPartiallySelected, hierarchical })
-    const tags = this.treeManager.getTags()
-    this.setState({ tree, tags })
+    this.initNewProps({ ...nextProps })
```
mrchief: Do you think this can be simply `this.initNewProps(nextProps)`?

src/tree-manager/flatten-tree.js
```diff
@@ -92,7 +92,7 @@ const tree = [
 }
 ```
  * @param  {[type]} tree              The incoming tree object
- * @param  {[bool]} simple            Whether its in Single slect mode (simple dropdown)
+ * @param  {[bool]} simple            Whether its in Single select mode (simple dropdown)
```
mrchief: 👍

src/index.js
```diff
@@ -226,7 +232,7 @@ class DropdownTreeSelect extends Component {
                   onAction={this.onAction}
                   onCheckboxChange={this.onCheckboxChange}
                   onNodeToggle={this.onNodeToggle}
-                  simpleSelect={this.props.simpleSelect}
+                  simpleSelect={isSimpleSelect}
```
mrchief: I feel that this could lead to potential issues down the line as it alters the meaning of `simpleSelect`. Up until now, `simpleSelect` lets you _safely_ assume certain things like no need to traverse below level 1, and the drop-down is a basically flat list. With this change, those assumptions are no longer valid and so it could become a potential source of trouble.

Thoughts?

src/tree-manager/tests/singleSelect.test.js
```diff
@@ -0,0 +1,35 @@
+import test from 'ava'
+import TreeManager from '..'
+
+test('should select the same node if it is selected and then deselected right again', t => {
```
mrchief: Not sure I understand this test. E.g., what is the need for this and why does it matter _only_ in the context of `singleSelect`? Can you help me understand, please?

src/tree-node/node-label.js
```diff
@@ -41,8 +41,8 @@ class NodeLabel extends PureComponent {
     const nodeLabelProps = { className: 'node-label' }
 
     // in case of simple select mode, there is no checkbox, so we need to handle the click via the node label
-    // but not if the control is in readOnly state
-    const shouldRegisterClickHandler = simpleSelect && !readOnly
+    // but not if the control is in readOnly or disabled state
+    const shouldRegisterClickHandler = simpleSelect && !readOnly && !disabled
```
mrchief: 👍

src/tree-node/index.js
```diff
@@ -29,7 +29,8 @@ const getNodeCx = props => {
       'match-in-children': keepTreeOnSearch && matchInChildren,
       'match-in-parent': keepTreeOnSearch && keepChildrenOnSearch && matchInParent,
       partial: showPartiallySelected && partial,
-      readOnly
+      readOnly,
+      checked
```
mrchief: Umm, why do we need this? As a courtesy?

ellinge: If you want to override the rendering of checked nodes. It's now currently difficult to do so by css since a selector which checks if a child is checked is not possible.

src/index.js
ellinge: This is like a hybrid which behaves like a simpleSelect but allows the traversing to be done. If you want it separated all the way we need to add checks and props everywhere where simpleSelect is used. The current solution allows it to behave as a simpleSelect everywhere except in the treemanager where it makes a difference.

src/tree-manager/tests/singleSelect.test.js
ellinge: Since singleSelect behaves differently than simpleSelect but is very similar this is to ensure we cover the lines for that.

ellinge:
@mrchief not sure what you mean with an additonal flag. Simpleselect should still behave as before. Did not want to break that so added singleSelect. If you provide both simpleSelect and singleSelect it will behave as before (no children)

Unfortunately did not fix it since that requires some more digging. Should be a separate issue.

ellinge:
> 
> 
> > I did not change the behaviour for simpleSelect so if that is something you think we should do instead, it would require additional changes.
> 
> Just one question - with this change, can we manage `simpleSelect` without the need of an additional flag?
> 
Not sure what you mean with an additonal flag. Simpleselect should still behave as before. Did not want to break that so added singleSelect. If you provide both simpleSelect and singleSelect it will behave as before (no children)

> > I also noticed on the storypage that simpleSelect is not working great with default values, which also applies to this solution. Only one default value should be possible and should be deselected when selecting something else
> 
> Great! Thanks for taking notice and fixing it!

Unfortunately did not fix the issue since that requires some more digging. Should be a separate issue.

src/index.js
mrchief: > The current solution allows it to behave as a simpleSelect everywhere except in the treemanager where it makes a difference.

For now, yes. And I see how this could have been misleading.

But let me look into this locally and I'll let you know.

src/tree-manager/tests/singleSelect.test.js
mrchief: Gotcha! In that case, I think we need to structure the test a bit different - checking/unchecking/checking seems like a generic enough task and not specific to single select. Yet, we're forced to do it here to meet code coverage requirements. I'd say those are probably not the right reasons to write a test. 

I'm not sure what the solution is yet, but I'd look into it to see if we can accomplish it in a sense that conveys the right purpose along with testing the required code paths. If not, we'll add a TODO here and move on for now.

src/index.js
```diff
@@ -55,11 +56,17 @@ class DropdownTreeSelect extends Component {
...
   }
 
-  createList = ({ data, simpleSelect, showPartiallySelected, hierarchical }) => {
-    this.treeManager = new TreeManager({ data, simpleSelect, showPartiallySelected, hierarchical })
+  createList = ({ data, simpleSelect, singleSelect, showPartiallySelected, hierarchical }) => {
+    this.treeManager = new TreeManager({ data, simpleSelect, singleSelect, showPartiallySelected, hierarchical })
     return this.treeManager.tree
   }
 
+  initNewProps = ({ data, simpleSelect, singleSelect, showPartiallySelected, hierarchical }) => {
```
mrchief: `createList` was my attempt at DRYing up the code but with `initNewProps`, we can inline that tree creation, something like:

```
  initNewProps = ({ data, simpleSelect, singleSelect, showPartiallySelected, hierarchical }) => {
    this.treeManager = new TreeManager({ data, simpleSelect, singleSelect, showPartiallySelected, hierarchical })   
    const tags = this.treeManager.getTags()
    this.setState({ tree: this.treeManager.tree, tags })
  }
```

src/index.js
mrchief: I looked into this a bit and I think we should separate it out. While separating it out results in more work on us, I think the benefits are worth the effort. E.g., right now, with `simpleSelect`, information is _lost_ in a sense that I can't tell whether it's simple or single. Secondly, we're adding CSS classes for easy targeting `className={cx('checkbox-item', { 'simple-select': simpleSelect })}` and if one wants to target simple differently than single, it is impossible to do so now.

Further, as I was looking into this, I realized, why are we making checkboxes act like radio buttons? Isn't rendering radio instead of checkbox semantically more correct? 

And what do you think about choosing a different name because `single` sounds too similar to `simple`. Maybe `radioSelect` or something else?

src/index.js
mrchief: `createList` was my attempt at DRYing up the code but with `initNewProps`, we can inline that tree creation, something like:

```suggestion
  initNewProps = ({ data, simpleSelect, singleSelect, showPartiallySelected, hierarchical }) => {
    this.treeManager = new TreeManager({ data, simpleSelect, singleSelect, showPartiallySelected, hierarchical })   
    const tags = this.treeManager.getTags()
    this.setState({ tree: this.treeManager.tree, tags })
  }
```

src/tree-manager/tests/singleSelect.test.js
mrchief: I guess this test will get rewritten anyway if we separate it out.

src/index.js
ellinge: I need the id set from the other PR to set a shared name (a.k.a the dropdown id) for the radio buttons.

ellinge:
@mrchief it's a radio select now. Fixed the issue with default values not getting unchecked as well + maximum amount of them getting selected for single select type dropdowns (first checked or isDefaultValue visited)

src/tree-manager/tests/singleSelect.test.js
ellinge: Added some more test and a snaphot now

ellinge:
@mrchief after some great expierence with code climate it's finally green and s a radio select now for review.
Fixed the issue with default values not getting unchecked as well + maximum amount (1) of them getting selected for single select type dropdowns (first checked or isDefaultValue visited)

ellinge:
@mrchief it's a radio select now. Fixed the issue with default values not getting unchecked as well + maximum amount (1) of them getting selected for single select type dropdowns (first checked or isDefaultValue visited)

src/index.js
```diff
@@ -53,11 +55,15 @@ class DropdownTreeSelect extends Component {
...
+    this.clientId = this.props.id
   }
 
-  createList = ({ data, simpleSelect, showPartiallySelected, hierarchical }) => {
-    this.treeManager = new TreeManager({ data, simpleSelect, showPartiallySelected, hierarchical })
-    return this.treeManager.tree
+  initNewProps = ({ data, simpleSelect, radioSelect, showPartiallySelected, hierarchical }) => {
+    this.treeManager = new TreeManager({ data, simpleSelect, radioSelect, showPartiallySelected, hierarchical })
+    const { tree } = this.treeManager
+    const tags = this.treeManager.getTags()
```
mrchief: The line above this popped an idea - if we implement a getter for `tags` in tree manager, then we can do:

```
  ...
  const { tree, tags } = this.treeManager
  ...
```

`getTags` is basically a result of years of programming in ES5. Don't have to make that change in this PR though. :)

src/tree-manager/flatten-tree.js
```diff
@@ -127,7 +136,8 @@ function setInitialStateProps(node, parent = {}) {
   }
 }
 
-function walkNodes({ nodes, list = new Map(), parent, depth = 0, simple, showPartialState, defaultValues = [], hierarchical }) {
+function walkNodes({
+  nodes, parent, depth = 0, simple, radio, showPartialState, hierarchical, returnValues = { list: new Map(), defaultNodes: [], checkedNodes: [] } }) {
```
mrchief: `returnValues` - interesting approach!

src/tree-manager/flatten-tree.js
```diff
@@ -165,13 +180,38 @@ function walkNodes({ nodes, list = new Map(), parent, depth = 0, simple, showPar
...
 
       node.children = undefined
     }
   })
-  return { list, defaultValues }
+
+  return returnValues
+}
+
+function ensureSingleSelect(defaultNodes, checkedNodes) {
```
mrchief: I think we can eliminate this (and the resulting traversal) if we keep track of selection while walking the tree. Something like (pseudo-code) 

```
if(radio && !defaultAlreadyChecked) 
  proceedWithCheckingDefaultValue
```

src/tree-manager/index.js
```diff
@@ -229,7 +249,7 @@ class TreeManager {
       if (node.checked) {
         tags.push(node)
 
-        if (!this.hierarchical) {
+        if (!this.hierarchical || this.radioSelect) {
```
mrchief: For radio, you're already keeping track of selected, so it'll be faster if we just return that instead of traversing the tree. Same for simple-select but simple select being a shorter list, the performance impact of traversing was negligible/minimal.

src/tree-manager/flatten-tree.js
ellinge: That would not work if a isDefaultValue is first and then a checked node comes along.

src/index.js
```diff
@@ -53,11 +55,15 @@ class DropdownTreeSelect extends Component {
       showDropdown: this.props.showDropdown || false,
       searchModeOn: false
     }
+    // This needs id generator from https://github.com/dowjones/react-dropdown-tree-select/pull/220
+    this.clientId = this.props.id
```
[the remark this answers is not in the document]
ellinge: Its in the other pull request. rootPrefixId is passed a long. A.k.a the id of the dropdown.

src/index.css
```diff
@@ -51,6 +51,18 @@
...
+      &.simple-select,
+      &.radio-select {
+        .checked .node-label,
+        .partial .node-label {
+          font-weight: 600;
+        }
+      }
+
+      &.simple-select .checkbox-item {
+        display: none;
```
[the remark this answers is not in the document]
ellinge: It’s more easily styled this way. Css is cascading so having the style deep down on a label is not something easily styled

src/index.js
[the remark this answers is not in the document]
ellinge: It should be passed through name on the checkbox component. If I have not missed that.

src/tree-manager/flatten-tree.js
ellinge: A bit odd but thought it would make sense to group the recursive props.
Perhaps should be prefixed with _?

src/checkbox/index.js
```diff
@@ -27,11 +28,11 @@ class Checkbox extends PureComponent {
   }
 
   render() {
-    const { checked, indeterminate = false, onChange, disabled, readOnly, ...rest } = this.props
+    const { checked, indeterminate = false, onChange, disabled, readOnly, radio, ...rest } = this.props
 
     const isDisabled = disabled || readOnly
 
-    return <input type="checkbox" ref={refUpdater({ checked, indeterminate })} onChange={this.handleChange} disabled={isDisabled} {...rest} />
+    return <input type={radio ? 'radio' : 'checkbox'} ref={refUpdater({ checked, indeterminate })} onChange={this.handleChange} disabled={isDisabled} {...rest} />
```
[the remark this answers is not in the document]
ellinge: Added a radio now. Thought about using const Tag = radioSelect ? RadioButton : CheckBox but left that idea. Also noticed some bug for firefox which don't seem to be an issue any more. The outside click is since then rewritten so I guess it's sorted out since then?

src/index.js
ellinge: Made some basic solution to this for now

src/tree-manager/index.js
ellinge: Done, returns tag for currentChecked

src/tree-manager/flatten-tree.js
mrchief: Definitely makes sense. No need for underscore though. I'm wondering if this is a code smell that is telling us that perhaps we should not use recursion or maybe actually use the returned values in the next recursive call? But that's for another day. Let's go with this for now.

ellinge:
Found an issue with IE11, it does not close the dropdown on single/simple selects. Develop-branch:
ie11bug

With lastest commit:
ie11fixed

ellinge:
Was an introduced issue with deselecting a tag.
This pr + 215 is available to check here:

src/index.css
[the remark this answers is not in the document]
mrchief: Unless it's fairly difficult to style the existing way, I would lean towards keeping the existing implementation. I'm not sure if anyone is relying on it so it's hard to tell. But if we're changing it, this needs to be called out and would constitute a breaking change. And, if we're gonna do it, I would do it in a separate PR with `BREAKING CHANGE: <rest of the title>` so that semantic-release can bump the versions accordingly. Again, if you think the change is worth going through all of that, then I'm up for it.

src/index.js
```diff
@@ -34,11 +34,13 @@ class DropdownTreeSelect extends Component {
     onFocus: PropTypes.func,
     onBlur: PropTypes.func,
     simpleSelect: PropTypes.bool,
+    radioSelect: PropTypes.bool,
     noMatchesText: PropTypes.string,
     showPartiallySelected: PropTypes.bool,
     disabled: PropTypes.bool,
     readOnly: PropTypes.bool,
-    hierarchical: PropTypes.bool
+    hierarchical: PropTypes.bool,
```
mrchief: We need to add trailing commas so that we can avoid noise in diffs like these. :)

src/index.js
```diff
@@ -53,11 +55,13 @@ class DropdownTreeSelect extends Component {
       showDropdown: this.props.showDropdown || false,
       searchModeOn: false
     }
+    // This needs id generator from https://github.com/dowjones/react-dropdown-tree-select/pull/220
+    this.clientId = this.props.id
```
mrchief: Can we take out `id` related changes? Those would come from 220 anyway and would simplify the diff ehre.

src/tree-manager/flatten-tree.js
```diff
@@ -92,31 +92,36 @@ const tree = [
...
+  const { list, defaultNodes, checkedNodes } = walkNodes({
     nodes: forest,
     simple,
+    radio,
     showPartialState,
     hierarchical
   })
-  return { list, defaultValues }
+  const defaultValues = defaultNodes.map(i => i._id)
+  const checkedValues = checkedNodes.map(i => i._id)
```
mrchief: Can we return the `ids` while walking so that we can avoid these 2 extra iterations?

src/tree-manager/flatten-tree.js
```diff
@@ -139,39 +147,57 @@ function walkNodes({ nodes, list = new Map(), parent, depth = 0, simple, showPar
       node._id = node.id || `${i}`
     }
 
-    if (node.isDefaultValue) {
-      defaultValues.push(node._id)
-      node.checked = true
+    if (node.checked) {
```
mrchief: I think we should not need this and this is why:

Consider this test:

```
test('should select single first checked node and ignore any defaultValues', t => {
  const tree = [{ id: 'nodeA', isDefaultValue: true }, { id: 'nodeB', checked: true }, { id: 'nodeC', checked: true }]

  const manager = new TreeManager({ data: tree, radioSelect: true })

  t.false(manager.getNodeById('nodeA').checked)
  t.true(manager.getNodeById('nodeB').checked)
  t.false(manager.getNodeById('nodeC').checked)
})
```
First of all, 

```
  [{ id: 'nodeA', isDefaultValue: true }, { id: 'nodeB', checked: true }, { id: 'nodeC', checked: true }]
``` 

is bad input. If it is radio select, one should not be passing multiple `checked`, but since it can happen, we can assume it will.

Next, `defaultValue` trumps other nodes. This is because `defaultValues` are checked by default and thus, `nodeA` is the first checked node. And then, we're following first-person-wins strategy here, so the first defaultValue should get checked and nothing else. Of course this all made up stuff but we need to use our best judgement and decide one startegy or the other. If we're going to stick to the aforementioned strategy, then what we're saying is:

 - `defaultValue` should be checked regardless of `checked` prop (and this rule is currenlty enforced so it's kinda non-negotiable...)
 - in case of radio select, we're checking the first checked node we find and ignoring the rest (this is new and thus can be changed much more easily)

So based on that, the assertion `t.false(manager.getNodeById('nodeA').checked)` is incorrect. The correct assertion would be:

```
  t.true(manager.getNodeById('nodeA').checked)
  t.false(manager.getNodeById('nodeB').checked)
  t.false(manager.getNodeById('nodeC').checked)
```

But then, things become complicated if someone passes data in a different order:

```
  [{ id: 'nodeA', checked: true }, { id: 'nodeB', isDefaultValue: true }, { id: 'nodeC', checked: true }]

  // correct assertions based on rules
  t.false(manager.getNodeById('nodeA').checked)
  t.true(manager.getNodeById('nodeB').checked)
  t.false(manager.getNodeById('nodeC').checked)
```

In this case, tree manager would set `nodeA` checked as well as `nodeB`. It should remove `nodeA` when it encounters `nodeB` and I think this is why you need the `if (node.checked) {...` check to counter balance all these scenarios.

Remember that tree-manger is in the critical path and a) it needs to be as lean and fast as possible; b) having too many conditionals/branches increases cognitive complexity

I'm wondering if we can avoid this extra conditional if tweak things a little; e.g. 

 - create separate `walkNodes` for `simple`, `radio` and `hierarchial`
 - use iteration instrad of recursion **_if_** it helps simplfying things by keeping track of checked nodes, default values etc.

Maybe do this for radio select to begin with and then change for simple/hierachial afterwards.

And if all of this leads to a less simpler solution than what you have currently, then leave your implementation as is.

If you prefer, I can play around with these a bit and if you checked "allow contributors edits" while creating this PR, I can push to this for your review. Let me know either way.

src/tree-manager/flatten-tree.js
ellinge: The reason I internally keep track of the nodes instead of ids is because I need to uncheck default values if finding a checked first. If I moved that to the return part it would be even more iterations depending on number of levels/nodes processed.

But you have different ideas about what should be checked first in the case of single select so perhaps that is not needed anymore.

src/index.js
ellinge: What do you mean, the comment or the line itself? It's needed for the rootPrefixId (inner ids)

src/index.js
ellinge: bild
You have a rule enabled preventing that

src/tree-manager/flatten-tree.js
ellinge: bild
You should have green light for adding changes.

ellinge: In my opinion isDefaultValue should not trump a checked-state. Not after reading "the dropdown will automatically select the node(s) when there is no other selected node". I think it implies it's a fallback value, a.k.a no checked.

Most of the checks are quite simple and starts with mostly the less expensive condition so not sure how much of a difference it would make performance wise.

As I said it should be up for editing.

src/index.css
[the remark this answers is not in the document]
ellinge: The only, simple way, now is to supply your own className to the component. Unless all your components use simple-select. See: https://stackoverflow.com/questions/2326499/apply-css-styles-to-an-element-depending-on-its-child-elements

I've restored the old, on checkbox-level, styling to avoid any breaking change but also supplied it on dropdownlevel for easier styling in the future. This also adds a lot of more bytes to the ouput since every checkbox needs to have the classname instead of just the dropdown.

src/index.js
ellinge: https://eslint.org/docs/rules/comma-dangle what should we target?

ellinge: https://github.com/airbnb/javascript/blob/b85baeafed8b66fdd9756439a0b8774860147913/packages/eslint-config-airbnb-base/rules/style.js

Goes with always-multiline for all types. Could simple be `comma-dangle: "always-multiline"` I guess.

src/tree-manager/flatten-tree.js
mrchief: Just trying to avoid extra iterations if possible.

src/tree-manager/flatten-tree.js
mrchief: > I think it implies it's a fallback value, a.k.a no checked.

Yes, but it still **_checks_** the input. Consider this - if someone provides `isDefaultValue: true, checked: false` - what would the checkbox do? Which prop do we honor (both on initial paint and after user actions of checking/unchecking other nodes)? 

This is what I meant by implied check - when we **init**, `isDefaultValue` implies it's checked and we ignore the checked prop.

src/tree-manager/flatten-tree.js
ellinge: I would have to add an if(depth === 0) and do the same iteration in walkNodes either way to avoid even more iterations. Unless we split upp the walking/traversing this is better since less checks are done in this way.

src/tree-manager/flatten-tree.js
ellinge: I've switched so isDefaultValue has precedence now. Have to anything regarding the iteration though.

src/tree-manager/flatten-tree.js
ellinge: I've added a benchmark here, it's very random "who wins" but I cannot see that the new iteration makes a big difference at least:
bild
bild

ellinge: IE11 is much slower though. Not sure if it's related to flattenTree or something else though:
bild

mrchief: Let me take a look and I'll get back to you.

I wouldn't worry about IE 11, the slowness is expected.

ellinge: After last commit it performs well in IE11 as well

src/index.css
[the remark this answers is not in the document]
mrchief: Can you show me some code examples of before/after using your approach? Just want to be sure I'm looking at the same thing that you are. 

> The checkbox-level targeting also adds more to the ouput since every checkbox needs to have the classname instead of just the dropdown.

GZip can take care of that (which is how I convince myself that the excessive div mania induced by React is OK).

However, I'm totally up for improving the status quo. I do think this merits its own discussion though, don't you? Breaking change or not, if you believe that we can simplify the existing CSS then let's go for it (not here as this is a big change already).

src/index.js
mrchief: I would go with `trailing-commas: "es5"` https://prettier.io/docs/en/options.html#trailing-commas

I started playing with a prettierrc in a separate branch but I ran into some weird prettier plugin errors. 

I would also do it after we merge all 3 of your PRs otherwise it'd just create a lot of extra work for you to update from develop branch.

src/index.js
mrchief: No, I mean `this.clientId = this.props.id` and anything related to id generation. If you need id changes here, then let's try and merge that PR first.

src/index.css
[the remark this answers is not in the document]
ellinge: You need to check the latest commit. This only shows the outdated css still.

Anyway if you want to style, e.g. simpleSelects labels differently than the regular tree and use both you won’t be able to. That’s unless you provide a class by yourself.
```jsx
<ReactDrop... className=”customsingle” />
```

```css
.react-dropdown-tree-select.customsingle {
    .dropdown-content label {
        color: blue;
    }
}
```

After adding the className on dropdown, for single and radio, you can now instead do (without any passed className):
```css
.react-dropdown-tree-select .simple-select {
    .dropdown-content {
        label {
            color: blue;
        }
        .checkbox-item {
            display: none;
        }
    }
}
```

There’s no possible way to style anything differently with only css without this additional className on the dropdown since you can’t write a selector targeting a parent (label) in css which depends of a child (.checkbox-item.simple-select). The only way is wrapping the component or passing a className which you can target.

With the addition it would make the additional checkbox className obsolete.

src/index.js
ellinge: I need the id for a shared name on radios. It still depends om the other pr.

ellinge:
This is... sooo stupid. Tried to ignore with filters without success. Tried to ignore with exlude_fingerprints but can't get a hold of the issue fingerprints...
We need to explore AST-ignores for this.props & this.state assignments and ignore of constructors etc.

mrchief:
Are you referring to code climate’s analysis? Don’t worry about it too much. :)

ellinge:
Yeah they are mentally wrong sometimes. Makes things worse by fixing them. We should explore the option of using filtering for some declarations and assignments we know can be ”false positives”. Such as const { .. } = this.props/state and constructor({ .. }). Any experience? Tried to understand it earlier today but didn’t get codeclimate in docker up and running. You can output the ast which could help you construct the filters.

https://github.com/codeclimate/codeclimate-duplication/blob/master/README.md#node-filtering

Used online ast-generator but failed to target the ”duplication”. Have had similiar issues with object deconstructions from props.

mrchief:
I had spoken to their support back when I first started using it. Took a bit to make it understand eslint 4 which made a lot of errors go away. I'm not too worried about them as long as they are known and acceptable. It's more of advice really and given the way React works, some of it is necessary. I did try to refactor it but gave up after the refactored code appeared more clunky. 

I can go into CodeClimate and ignore an issue if it's a known one or a reasonable one. I've done it for past PRs. Never thought of delving deeper into their code nor do I have the time for it honestly.

.codeclimate.yml
```diff
@@ -34,6 +34,12 @@ plugins:
   eslint:
     enabled: true
     channel: "eslint-4"
+  duplication:
```
mrchief: Does it help? As I said, I can go and ignore warnings for existing code and any new code which gets triggered should be reviewed (and can be ignored if it's a false positive). My concern is this would mask the initial trigger.

ellinge: Yes. Helped in this case. Default threshold is 45.
https://docs.codeclimate.com/docs/default-analysis-configuration

mrchief: In that case, this is fine!

ellinge:
Discussed in other PR, added option to keep dropdown open for radio/simple

src/tree-manager/flatten-tree.js
```diff
@@ -141,39 +146,58 @@ function walkNodes({ nodes, list = new Map(), parent, depth = 0, simple, showPar
       node._id = node.id || `${rootPrefixId ? `${rootPrefixId}-${i}` : i}`
     }
 
-    if (node.isDefaultValue) {
-      defaultValues.push(node._id)
+    if (single && node.checked) {
```
mrchief: I tried a few different ways to see if we can somehow make this any simpler. With our current "First Person Wins" and "`defaultValue` supersedes `checked`" strategy, things end up being complicated since we need to backtrack.

As we walk the forest/tree, we are moving forward, so things become a lot simple if we switch to "Last Person Wins" strategy. 

Something like: 

```
  walk 
    if isDefaultValue
        set node checked

    if node is checked 
       if single
          checkNodes[0].checked = false
          // optionally throw a warning to the user maybe?? Feel free to change the wording if you like...
          if checkNodes[0].isDefaultValue
             console.warn("You've assigned more than 1 default values while the component is in single select mode (you've set either `simpleSelect=true` or `radioSelect=true`). Only the last node will be honored. If you intend to use multiple values, switch to multi-select instead.")
              // not sure if we should clean up the input here by setting isDefaultValue to false
          
          // update with latest node since Last Person Wins
          checkNodes = [node]
          
       else 
          checkNodes.push(node)
```

It _should_ bring down the Cognitive complexity score as well. I as too exhausted to try this last night so I'm not sure if this will work 100% as expected but worth a try IMHO. Let me know your thoughts.

ellinge: Give it a go. Won't you have problem distinguishing a previous checked node which was done by isDefaultValue then though? If isDefaultValue should no longer supersede checked, or vice versa,  then the current logic would be simplified also.

mrchief: This is just a convenience as most of the issues can be resolved by setting the proper data, so I don't want to go crazy on this. Certainly not worth sacrificing performance and increasing the complexity.

We can simply go with "last person wins" or "first person wins" with whatever takes the simplest of solutions. 

> If isDefaultValue should no longer supersede checked, or vice versa, then the current, "First person wins" logic could be simplified also.

With you. In order to both simplify logic and make sure we're clear about the behavior, what do you say to this?

(only for **single** select cases since this is a non-issue for multi-select )

 1. Setting default values - First `isDefaultValue` wins (helps us shortcircuit the tree iteration)
 2. Setting checked values - Last person wins (we walk all nodes anyway and is easier to override previous values) don't know if we wanna go with a universal theme of FPS or LPS in all cases
 3. When both defaultValues/checked present - treat defaultValue nodes as checked and then follow `2`

Will try this and see where I land. Just thinking out loud here so I'm sure I'll find something while implementing and testing.

mrchief: I tried this and I don't see it being any simpler. I think what we have right now is pretty good. Unless you have any objections, this is good to be merged.

ellinge:
I bundled the selection of rendering / behaviour mode into a single prop instead like we discussed in the keyboard nav PR. We can fill the migration steps with additional breaking changes (for on action) and if bundle the text-props also.

docs/guides/migrating-from-1-to-2.md
```diff
@@ -0,0 +1,9 @@
+## React Dropdown Tree Select Migration Guide v1 to v2
```
mrchief: We shouldn't be needing this for radio select. It's a new addition that's backward compatible.

mrchief:
> I added the textprops-bundle in this pr as well (without the new labels). With the onaction-changes we would cover all breaking changes needed for a release I think if we want to wait with ARIA/kryboard nav.

Will it be too much to take out the bundle props out of this?

I think radioSelect, ARIA, keyboard nav are all backward compatible changes so we can merge them first and release them incrementally. onActions and bundle props then become part of the v2 release.

This release sequence lets v1 folks to not miss out on most new features. Upgrading to major versions is generally considered a bigger undertaking and makes people hesitant in upgrading which would cause them to miss out on accessibility and radio select for no good reason. 

I'm open to arguments otherwise.

ellinge:
Well since you [became reluctat with the prop additions](https://github.com/dowjones/react-dropdown-tree-select/pull/225#discussion_r275099089) I thought it would be to first add them and then in the next release move them. I see all these features as mayor additions and the migration steps simple enough to not scare any people away. Simple select and noaction could be 2.0 and wcag 2.01. But I can revert them...

mrchief:
These are major additions, no doubt about that. But not everything is a breaking change. Another reason for releasing them separately is backward compatibility. Id-generation logic, Radio Select, ARIA, Keyboard are all backward compatible. Releasing them as v1 will make it a frictionless upgrade for anyone. And upgrading to 2.0 is not a big hurdle but it still requires _some_ changes. 

You've done a great job at sending them as separate PRs and I'd like to take advantage of that fact. I can help with this process as well.

docs/guides/migrating-from-1-to-2.md
ellinge: Reverted

README.md
```diff
@@ -49,7 +49,8 @@ A lightweight and fast control to render a select component that can display hie
   - [noMatchesText](#nomatchestext)
   - [keepTreeOnSearch](#keeptreeonsearch)
   - [keepChildrenOnSearch](#keepchildrenonsearch)
-  - [simpleSelect](#simpleselect)
+  - [keepOpenOnSelect](#keepopenonselect)
+  - [mode](#mode)
```
mrchief: This is a breaking change and most likely belongs to bundling props PR. I thought we're merging this without needing any breaking changes?

mrchief:
@ellinge Thanks for the quick update. You're the best! I'm gonna merge this unless you have any other changes to make. Let me know.

PR #218, opened 2019-03-13
fix: Do not allow delete of tags on disabled dropdown

### LINKED ISSUE #202: User can delete preselected tags even when control is disabled
**Describe the bug**
If you initialize the tree with pre-checked values and set `disabled=true`, the control still lets users remove the tags.

**To Reproduce**
 - Go to https://dowjones.github.io/react-dropdown-tree-select/#/story/options
 - Select `disabled` checkbox
 - Click on `x` on the preselected tag

**Expected behavior**
The control should not allow deletions.

**Screenshots**
mnfopvw8wx

ellinge:
Fixes #202

- [x] Bug fix


mrchief:
@ellinge You're killing it! Thanks for your PRs. They look fine at a glance. Let me know once you're done with your changes and I'll review them.

Let's add @ellinge for code, ideas cc @allcontributors[bot]

ellinge:
@mrchief I think this is one is up for grabs. Reused the readonly-prop with a fallback on disabled and added a test

src/tag/index.test.js
```diff
@@ -33,6 +33,13 @@ test('call onDelete handler when pill is closed', t => {
   t.true(onDelete.calledWith('abc'))
 })
 
+test('should not call onDelete when readOnly', t => {
```
mrchief: We should add a test for `disabled` as well.

mrchief:
I think we shouldn't reuse the `readOnly` prop. That prop also gets passed down and `readOnly` and `disabled` are different states with different meanings in HTML world. Other than that, this looks good.

src/tag/index.js
```diff
@@ -22,12 +23,12 @@ class Tag extends PureComponent {
...
 
   render() {
-    const { label, readOnly } = this.props
+    const { label, readOnly, disabled } = this.props
 
     return (
       <span className={cx('tag')}>
         {label}
-        <button onClick={!readOnly && this.handleClick} className={cx('tag-remove', { readOnly })} type="button">
+        <button onClick={!readOnly && !disabled ? this.handleClick : undefined} className={cx('tag-remove', { readOnly }, { disabled })} type="button">
```
mrchief: Why the explicit `undefined`? I mean React already takes care of it if you do `(!readOnly && !disabled) && this.handleClick` or `!(readOnly || disabled) && this.handleClick`

ellinge: The other expression triggered a warning in the tests. "If you used to conditionally omit it with onClick...". I can change it back.

ellinge: Perhaps related to https://github.com/alexkuz/react-json-tree/pull/97
Not sure if that's in the dependency graph of ours right now.

ellinge: Also se https://stackoverflow.com/a/54038930

mrchief: Well, that's annoying.

PR #220, opened 2019-03-14
feat: Add unique ids on dropdowns if not provided

### LINKED ISSUE #208: DropDownTree selection can be broken when using multiple instances on the same component
**Describe the bug**
When using two or more DropDownTree components in the same page, sometimes clicks for selecting a dropdown do not unselect the previously selected dropdown.

**To Reproduce**
https://codesandbox.io/s/j2z20j56rw

1. click on the first dropdown
2. select an element
3. click on the second dropdown

Result: Both dropdowns are open

**Expected behavior**

There should only be only one dropdown selected or opened at a time

**Screenshots**

image

Reproducible on any browser

**Additional context**
It seems like the function `utils/isOutsideClick` can detect if the dropdowns are different only if the `className` property is different, maybe an additional generated key by component instance could fix this method.

### LINKED ISSUE #212: Different behavior (selecting the checkbox VS clicking the label) with multiple instances on the same page

### LINKED ISSUE #215: Generate node id that is unique across multiple instances
In order to correct name label/inputs, we need to generate unique id/names. The current solution of generating node id guarantees unqiueness only within the same instance. It fails when multiple instances of the component are created on the same page.

ellinge:
Added possibility to provide a id by property. If not provided it generates a unique by a global counter.

Fixes #215
Fixes #212
Fixes #208 

## Type of change

- [x] Bug fix
- [x] New feature


ellinge:
Thinking that we also could solve it by passing the domnode for the react component (by storing ref) when checking the event hierarchy instead of comparing attributes (className id). No need for id then at all.

ellinge:
But if we need unique ids on the checkboxes as well the generated ids could be used as a prefix for them. Tried that but getNodeById etc is dependent on the _id attribute of the node so no match in that case. The tree manager would then have to remove added prefix when looking up nodes.

mrchief:
Oh boy! You're on 🔥 

> Thinking that we also could solve it by passing the domnode for the react component (by storing ref) when checking the event hierarchy instead of comparing attributes (className id). No need for id then at all.

I think we should explore this idea (in a separate branch maybe?) and see which one fares better.

ellinge:
Already in progress, seems to work great with this.node-check instead. Fixed so that children are prefixed with the dropdown id. Coming up soon.

src/tree-manager/flatten-tree.js
```diff
@@ -94,17 +94,19 @@ const tree = [
  * @param  {[type]} tree              The incoming tree object
  * @param  {[bool]} simple            Whether its in Single slect mode (simple dropdown)
  * @param  {[bool]} showPartialState  Whether to show partially checked state
+ * @param  {[string]} rootPrefixId    The prefix to use when setting root node ids
  * @return {object}                   The flattened list
  */
-function flattenTree(tree, simple, showPartialState, hierarchical) {
+function flattenTree(tree, simple, showPartialState, hierarchical, rootPrefixId) {
```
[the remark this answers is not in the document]
ellinge: Not sure what to do here @mrchief 
Any thoughts?

README.md
```diff
@@ -309,6 +310,14 @@ Type: `bool` (default: `false`)
 `disabled=true` disables the dropdown completely. This is useful during form submit events.
 `readOnly=true` makes the dropdown read only, which means that the user can still interact with it but cannot change any of its values. This can be useful for display only forms.
 
+### id
+
+Type: `string`
+
+Specific id for container. The container renders with a default id of `rddts-N` where N is count of the current component rendered.
```
mrchief: dropdown is a single word, so maybe this should be `rdts`?

More on the `-N` suffix below :)

__snapshots__/src/index.test.js.md
```diff
@@ -171,110 +175,110 @@ Generated by [AVA](https://ava.li).
...
                   children: undefined,
                   label: 'item1-1',
                   value: 'value1-1',
                 },
-                '0-0-0' => {
+                'rddts-0-0-0' => {
                   _depth: 2,
-                  _id: '0-0-0',
-                  _parent: '0-0',
+                  _id: 'rddts-0-0-0',
```
mrchief: The nodes are labeled as `[parentId-]depth[-arrayIndex]`. I didn't have a strong reason for doing so at the time but it felt like an easy enough thing to do and gives us some kind of information encoding for each node that doesn't have a preset id. It's borderline silly I guess but I'm not sure if anyone is relying on this pattern or not. The `-N` suffix throws off this scheme. We can simply have the id as `rdtsN` or `rdts.N` or something other than the `-` and the existing scheme would still be intact. Thoughts?

src/index.test.js
```diff
@@ -148,3 +162,59 @@ test('deactivates dropdown active on blur', t => {
   wrapper.instance().onInputBlur()
   t.false(wrapper.instance().keepDropdownActive)
 })
+
+test('detects click outside', t => {
```
mrchief: Thanks for adding this, was much needed! 🙏

ellinge: Np. Was not easily fixed without adding any new dependency since dispatchEvent/simulate does not seem to bubble to the virtual jsdom document eventlistener. Had trigger it manually. Maybe could be fixed better if adding simulant.js.

__snapshots__/src/index.test.js.md
ellinge: Sure, I can add a separator prop when constructing the generator which we use blank or dot for.

__snapshots__/src/tree/index.test.js.md
```diff
@@ -1,13 +0,0 @@
-# Snapshot report for `src\tree\index.test.js`
```
mrchief: Do you know why this file is showing up as deleted? Based on the changes, there should have been no changes to this snapshot.

src/index.test.js
```diff
@@ -53,44 +56,44 @@ test.beforeEach(t => {
 
 test('renders default state', t => {
   const { tree } = t.context
-  const wrapper = mount(<DropdownTreeSelect data={tree} />)
+  const wrapper = mount(<DropdownTreeSelect id={dropdownId} data={tree} />)
```
mrchief: I don't think we need to update all the existing tests with id. To ensure that everything is backward compatible, we need to make sure that not passing the new prop is not breaking anything. It should autogenerate one (id), if not passed, correct? Correct me if I'm wrong here, but I'd imagine that the existing tests will have autogenerated id in their snapshot, and an additional test that passes `id` to ensure correctness of the new implementation.

src/tree-manager/flatten-tree.js
```diff
@@ -138,6 +140,9 @@ function walkNodes({ nodes, list = new Map(), parent, depth = 0, simple, showPar
     } else {
       node._id = node.id || `${i}`
     }
+    if (depth === 0 && rootPrefixId && rootPrefixId.length > 0) {
```
mrchief: `rootPrefixId` is a string and the beauty of the language is that just the first check would suffice. I would imagine that the length check may be needed for cases where it's all whitespaces (so you'd trim it and then check for length or something like that). 

However, I think we don't need this extra `if`. It can be covered in the `else` branch at L141

```
} else {
      node._id = node.id || `${rootPrefixId ? `${rootPrefixId}-${i}` : i}`
```

The logic is: 
 - use `node.id` if explicitly passed (this is current contract)
 - otherwise, create our own:
   - prefix `rootPrefixId` if passed
   - otherwise, just use the counter

src/tree-node/actions.test.js
```diff
@@ -15,12 +15,12 @@ test('renders actions', t => {
     }
   ]
 
-  const wrapper = toJson(shallow(<Actions actions={actions} />))
+  const wrapper = toJson(shallow(<Actions actions={actions} id="snapshot" />))
```
mrchief: What's this `id` used for? `Actions` is not a component by itself, just a renderer for an array of actions. It should not have any props of its own unless we need it for something...?

src/utils/generateUID.js
```diff
@@ -0,0 +1,22 @@
+export const generateUID = (prefix, separator) => {
+  let counter = 1
+
+  let map = new WeakMap()
+
+  const get = item => {
+    if (!map.has(item)) {
+      map.set(item, counter++)
+      return get(item)
```
mrchief: I guess we can remove the extra function call here? Without the return, it'll execute the same fall through logic if I'm not mistaken.

src/utils/generateUID.js
```diff
@@ -0,0 +1,22 @@
...
+    }
+    return `${prefix || 'uid'}${separator}${map.get(item)}`
```
mrchief: Would it be better if we use the array notation - like `rdts[0]`, `rdts[1]` and so on? I'm wondering if it'd help in a form submit. In that case, for a single instance, we would have to emit just `rdts`. We wouldn't know if there are multiple instances ahead of time so maybe this is a problem that the user needs to solve.

src/utils/generateUID.js
```diff
@@ -0,0 +1,22 @@
...
+
+  const reset = () => {
+    map = new WeakMap()
+    counter = 1
+  }
+
+  return { get, reset }
+}
+
+export const rdtsUID= generateUID('rdts', '')
```
mrchief: Unless we plan to make it a generic utility method usable outside of the component, I guess we can tighten this up a bit more.

E.g.
 - `generateUID` is not used anywhere outside of this file, so we can choose not to export it
 - eliminate branches like `prefix || 'uid'`
 - eliminate optional separator logic - decide if we want to have a separator or not and then design the API accordingly.

Next, about its API surface; imagine someone **new** looking at `this.clientId = props.id || rdtsUID.get(this)` and trying to understand the code. 

Is `rdtsUID.get()` self-descriptive enough or can we do more - so that one doesn't need to peek at its implementation to figure out what it means or does? How about something like `getInstanceId.next(this)` or `nextUniqueId.get(this)` or `clientIdGenerator.get(this)` or similar? 

If one is curious about _how_ it's implemented, they would still need to peek at the implementation, but if I'm just trying to understand the overall flow, maybe this tells me enough about what the intended op is?

src/utils/isOutsideClick.js
```diff
@@ -12,8 +12,7 @@ const getPath = e => {
   return path
 }
 
-export default (e, className) => {
+export default (e, node) => {
   if (!(e instanceof Event)) return false
-  const completeClassName = className ? `${className} react-dropdown-tree-select` : 'react-dropdown-tree-select'
-  return !getPath(e).some(node => node.className && node.className.indexOf(completeClassName) >= 0)
+  return !getPath(e).some(eventNode => eventNode === node)
```
mrchief: This is what `node.contains` was supposed to do but the native method was not reliable across browsers. We had issues in firefox which is why `isOutsideClick` was born. Hopefully, this implementation will fare better.

src/tree-node/actions.test.js
ellinge: It's still a required prop so causes a warning in tests unless supplied.

src/index.test.js
ellinge: I've updated all snapshots with id to not rely on order they're executed. There's a separate test to ensure unique ids.

src/utils/isOutsideClick.js
ellinge: Did you check on the node itself also? Or what was the problem? Just to ensure it's not introduced again. I've only tested it with latest ff, chrome, edge and ie11.

mrchief: This is one of the early attempts that caused issues on Firefox: 

https://github.com/dowjones/react-dropdown-tree-select/blob/bbd5c2ca62cb4f563dc49c8a7bcf680fa872370f/src/index.js#L89-L95

__snapshots__/src/tree/index.test.js.md
ellinge: You've  or someone has set skip on the snapshot-tests for related file meaning no snapshots are generated. Was not aware of the --update-snapshots command at first so just deleted and checked changes. I let them be deleted since I saw that they're were skipped.

src/utils/isOutsideClick.js
ellinge: Ok not sure why that would not work. Any steps to reproduce the issue to verify it's not reintroduced?

src/utils/generateUID.js
ellinge: I agreed/landed with clientIdGenerator

src/utils/generateUID.js
ellinge: I don't think it would be simple to detect single instances. For form submits we need to ensure the rootprefix/clientid is used as name for checkboxes as well if they should be grouped. I've passed the clientId down to the name in the latest commit.

ellinge: I don't think you're should use brackets in ids. Don't have any source for it now but I think Asp. NET MVC does some other naming when creating arrays of objects for submission. We can however append brackets on the name for checkboxes to group them in form submission.

ellinge: This is the reason https://www.w3.org/TR/html401/types.html#type-name
But that’s removed in html5 though.
Not sure how you would use a css selector for attributed on a id-tag with brackets though, even if an unlikely use case.

What about keeping the current format but adding brackets for the name (using clientid) of checkboxes to post them as an array? See example here:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#Using_checkbox_inputs

Then users could supply their own id, such as 'interests', in form submissions.

src/tree-node/actions.test.js
ellinge: Warning: Failed prop type: The prop `id` is marked as required in `Actions`, but its value is `undefined`.
    in Actions

src/utils/generateUID.js
ellinge: Latest commits are up and running here: https://ellinge.github.io/react-dropdown-tree-select-test/

src/utils/isOutsideClick.js
mrchief: If it works on the latest FF, then I think it's a good enough indicator that it fixes the issue. Let's roll with that for now.

src/utils/generateUID.js
mrchief: I love the test page! If you don't mind, I think it should be included as a story on the docs site. 

Looking at the radio select, I think we should keep the dropdown open after clicking on a radio. This would help in accidental clicks in a long list and also would feel more in line with current behavior.

ellinge: About the test page, perhaps it's better with something like the Options-story? There’s a lot of dropdowns… The only reason I created it was to test it with IE11 properly. But I won’t mind if you want to use it.

About the radio select, what about an option at least to close on select or the other way around, to keep it open? We use the radio select with hidden inputs (similar to simple select) and as an tree in some cases for multi selects and are happy with its current state.

mrchief: > About the test page, perhaps it's better with something like the Options-story? There’s a lot of dropdowns… 

Yeah, I agree. We can think about it later. 

> About the radio select, what about an option at least to close on select or the other way around, to keep it open?

Sure, that's a good idea.

src/tree-node/node-label.js
```diff
@@ -51,7 +53,7 @@ class NodeLabel extends PureComponent {
     return (
       <label title={title || label} htmlFor={id}>
         <Checkbox
-          name={id}
+          name={`${clientId}[]`}
```
mrchief: Did you mean to keep the square brackets here?

ellinge: Yes, generating dropdownid[] as a name means all checkboxes gets handled as an array om form submits. At least for most server side receivers such as .NET MVC.

ellinge: Se https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#Using_checkbox_inputs

Not sure how form submits are handled by the component even since they probably aren’t even rendered on the client on form submit? But either way that should be the way to handle grouping of releated checkboxes.

mrchief: Are you using it as such? I never intended the checkboxes to constitute the form values - mostly using the events to dispatch redux actions has been my primary use case. Interesting idea though...

mrchief: I know about the array notation, just that I never put much thought into making it act as a form control. I've used it with redux-form mostly where the flow I've used is `node change > onChange > dispatch (redux-form manages this) > submit (redux-form handles the form values`. I've also used it with just Redux but the flow is mostly the same - `onChange > dispatch > use store to submit values`.

Ok, so that aside, I think this can be merged (unless you're planning on making any more changes?).

mrchief: Oh, remembering the array notation and since you mentioned MVC - in order for serialization/deserialization to work properly, we'd need to render it without array notation in case of a single checkbox (with array notation it'll always be deserialized as an array and would thus fail if the receiving prop is not an array).

Also, a plain form submit may not be desirable since one may not want to submit all checkboxes all the time - e.g, with hierarchial mode, one may want to submit only the parent node if all children are selected.

If you need this feature, then let's open up a separate thread for this. Otherwise, let's stick to just ids for now. I want to start merging the PRs :)

ellinge: Actually it was based of your idea: https://github.com/dowjones/react-dropdown-tree-select/pull/220#discussion_r268414543

Do you want the unique node id back or the clientid without brackets? Current dev has a unique name per checkbox.

If we wanted the form submit to work we could just introduce a hidden field per tag with this suggested bracket naming and it should work with a regular form submit. Not something we use currently though. So this was all based on you suggestion.

ellinge: The old naming is restored, a.k.a no change on name for checkboxes

ellinge: So up for merge.

mrchief: > Actually it was based of your idea:

You missed the "I'm wondering..." part but I hear ya. My bad!

> Current dev has a unique name per checkbox.

Only works for single instances as you can see with #212.

> Do you want the unique node id back or the clientid without brackets?

Both actually! To fix #212, we need to make the checkbox id unique across instances. So I guess something like `${clientId}-${id}` should do the trick?

ellinge: The id is unique. The name is unique. It will work. No change is needed. The name is related to form submit only. Which web currently don’t support native I guess.

PR #223, opened 2019-03-21
fix: Add TypeScript definitions

### LINKED ISSUE #209: TypeScript bindings
This is a great library. However, it would be even better if it included TypeScript bindings, or at least third-party bindings via https://github.com/DefinitelyTyped/DefinitelyTyped .

For now I've created a `types/react-dropdown-tree-select.d.ts` file which contains the following:

```typescript
declare module 'react-dropdown-tree-select' {
  export default class DropdownTreeSelect extends React.Component<any, any> {}
}
```

ellinge:
Adds TypeScript-definitions

I also noticed that onAction is not correctly implemented. Currently sends a { action: string, id: string }.
There is code in index.js which tries to fetch the node. But it does not seem be working. It expects two parameters but receives the object described above in the first parameter.
bild

Changed the examples/stories to get the node id at least.
Also changed in readme where onAction should be applied.

Fixes #209 

- [X] New feature


ellinge:
Tested with npm install git:
bild

Currently based on develop so radioSelect (and id is missing).

mrchief:
@ellinge This is really great!

mrchief:
Are you done with this? It seems like we can merge this and get a release out (will get radio once #217 gets merged).

README.md
```diff
@@ -149,8 +149,8 @@ const data = {
 const onChange = (currentNode, selectedNodes) => {
   console.log('onChange::', currentNode, selectedNodes)
 }
-const onAction = ({ action, node }) => {
-  console.log(`onAction:: [${action}]`, node)
+const onAction = ({ action, id }) => {
```
mrchief: I think this highlights an issue with documentation. 

`{ action, id }` is basically `actionData` and it can be anything the caller wants to propagate. In my examples, I chose `action` and `node` for no particular reason other than the fact that I just wanted to demo something.

Similarly, each `action` object can have it's own `onAction` handler instead of one single `onAction`. Or you can have a single/common `onAction` like the readme shows. 

I believe having individual `onAction` allows for greater flexibility. It also frees you from declaring a big `onAction` handler with multiple if-else/switch branches for each action id but if that suits one's use case better, they can choose to do so.

The common `onAction` has a signature of `func(action, node)` and not `func({action, node})`; so that's a documentation bug.

The individual action's `onAction` takes precedence over the common `onAction` if passed via the ES6 rest/spread/merge magic here:

https://github.com/dowjones/react-dropdown-tree-select/blob/77e46eebb0406008afb5f352c1a044d84fb53f5b/src/tree-node/actions.js#L14-L20

`...rest` contains the common `onAction` which will get overridden by individual action's `onAction` via `{...rest} {...a}` statements. 

We need to do a better job at pointing this out I guess. But I don't see any bugs in the code per se (readme, stories yes, but the code should be working as intentded). Let me know if I missed something.

ellinge: Actually tried to pass it by creating a specific handler on an action. Got an error at first so I understood it probably got overriden by the spread. When passing a correct handler I still got the global events output so thought it was an error in the documentation only. 

From what I can se only the global onAction tries to fetch the node. But since nodeid is not passed to the event it fails.

The actionData does not contain the action object imho as hinted by the readme. It should be the complete action as defined with perhaps onAction removed. I can't see where that magic would happen but it was late so perhaps overlooked something.

What should the global action and node action event have as declaration so that it matches the current behavior?

ellinge: I've created different scenarios here:

It's my code base but have not touched anything regarding actions so should not be an issue.
I can only get it to work with a global onAction. And the only thing working is the declared actionData, a.k.a { action: actionId, id }, a.k.a { actionId, nodeId }

Code running is here:
https://github.com/ellinge/react-dropdown-tree-select-test/blob/master/src/App.tsx

What have I misunderstood? I can't really wrap my head around how it could work when checking the source.

It's important that the TypeScript matches the actual implementation. Therefore the documentation and declartion should match it in this PR and we fix the actual problem in a separate issue.

ellinge:
> 
> 
> Are you done with this? It seems like we can merge this and get a release out (will get radio once #217 gets merged).
> 
> Also, I think we should separate out action related changes (see my comments regarding them) so that we can get the typings out.

I want to be sure with the typing for onAction at first. Also I think need to split treenode as well since the return-object does not include children anymore (since it's set to undefined). I should also add the generated id on the return node and action as well.

Do you think we should expose any "internal" props as well. Such as node and searchInput. We actually use them to set the width of the dropdown to the div-node. I guess others are interested in their expose as well without casting the whole ref to any instead.

README.md
mrchief: Let me take a look and I'll get back to you.

mrchief:
> Should we expose the _id/id props as well?

No, since they are internal to the component. If anyone needs `id`, they can pass their own and the component will use them. I don't want anyone relying on our internal mechanism of generating IDs.

> Do you think we should expose any "internal" props and methods as well. Such as node and searchInput. We actually use them to set the width of the dropdown to the div-node and handling tabbing in and out of the dropdown. I guess others are interested in their expose as well without casting the whole ref to any instead.

Let me think about this and I'll get back to you.

README.md
ellinge: Great, I updated the code to skip using bigdata, It's a bit faster now without needing to parse the json * 3 on each click.

mrchief: This [commit](https://github.com/dowjones/react-dropdown-tree-select/pull/35/files#diff-2555cbaa60f72d8b5386e400a8c9297aR7) broke things when I removed `cloneDeep`. 

If you remove the clone part and replace it with `const { list, defaultValues } = flattenTree(data, simpleSelect, showPartiallySelected, hierarchical)` at [tree-manager/index.js#L9](https://github.com/dowjones/react-dropdown-tree-select/blob/develop/src/tree-manager/index.js#L9), you can see things working as intended:

e8HLYZop68

This is a bug and lack of tests let it slip through undetected 😞.

We can remove cloning entirely or bring back `cloneDeep` or something lighter if we need cloning. Honestly, I'm not sure if cloning is required or if adds any value or if it's something that's beneficial to the end caller. The sole reason I did it was to not modify inputs. But we can achieve the same objective by doing a new node and copying existing node props to it during `walkNode`. Anyway, I'll a bug for this and  fix it, this time with tests for actions so that future slip ups do not happen.

ellinge: Alright so the cloning caused the "local" onClick to fail? Sounds reasonable. And the solution sounds correct but still needs some generic cloning since you allow the user to pass custom data. JSON.stringify does not handle handlers/func but could still be an option if setting onClick explicitly after. That's unless the user wants to supply other custom functions, which they in theory could? 

What about the return values for now though. Looks like you also get { action: string, id: string } returned. So the typings and documentation currently matches the implementation? So this should br a go and adjusted in the fix for onAction?

mrchief: Yep! The return values are consistent. What I called `node` in docs is actually the node id, so changing it to `id` seems better.

If I remember correctly, the global `onAction` was introduced later as a feature request. I'm going to add an issue to add documentation and a few use case examples around `onAction` to make it more prominent.

ellinge: Would it not be better to actually return the action and node objects as it said for the current onAction documention? The documentation also specifies that you can supply additional properties which is returned in onChange events. Shouldn’t that be consitent with the onAction-event? The global action trigger has been prepared for this it seems but does not get passed the required nodeid to make it happen. That would require that the local event to get handled in the same way though.

mrchief: > Would it not be better to actually return the action and node objects as it said for the current onAction documentation?

`node` is an internal object, an implementation detail. I don't want anyone to rely on it as that limits our ability to change it freely.

> The documentation also specifies that you can supply additional properties which is returned in onChange events. 

For the node's `onAction`, yes. And I think it does pass it in the event handler.

> Shouldn’t that be consistent with the onAction-event? The global action trigger has been prepared for this it seems, but does not get passed the required nodeid parameter to make it happen. That would require that the local event to get handled in the same way though.

I'm not sure about this. Would actually like to see it in code before I can make a call. I'm also not sure how much feature parity we want to impose on local vs global `onAction`. These are good topics for discussion.

ellinge: The internal node is already passed for the onChange event. It's the reason you can access your custom properties. You can access _id if you want also. 
https://github.com/dowjones/react-dropdown-tree-select/blob/e5eb14035d642d9734ba5f80d00eda66456b311c/src/index.js#L168

Nothing except the action's id and node's id gets passed in the event handler. I think we should pass the whole action object and the node object (such as for onChange) as said in the current documentation.

If you ignore the fact that local onAction isn't working these are the lines why nothing except ids gets passed in the event handler:
- Creation of actionData (only actionid and nodeid) 
https://github.com/dowjones/react-dropdown-tree-select/blob/e5eb14035d642d9734ba5f80d00eda66456b311c/src/tree-node/actions.js#L20

- Triggering of event (passes actionData) 
https://github.com/dowjones/react-dropdown-tree-select/blob/e5eb14035d642d9734ba5f80d00eda66456b311c/src/tree-node/action.js#L19

- Global trigger of event (receives actionData and passes it forward and tries to lookup node from missing parameter) 
https://github.com/dowjones/react-dropdown-tree-select/blob/e5eb14035d642d9734ba5f80d00eda66456b311c/src/index.js#L172

ellinge: I've added a dropdown selection for rendering 1.16.0 vs developTemp here:

As you can see onAction does not contain any more information.
Even if I try to fetch an additional parameter.
`  const onAction = (action: any, node: any) => 
    addToLog("onAction", { action: action || null, node: node || null });`

mrchief: > The internal node is already passed for the onChange event. It's the reason you can access your custom properties. You can access _id if you want also.

Only because I'm not filtering out internal props. But it's not a guarantee. The only guarantees are the ones provided in the docs. It's not strictly enforced at the moment because I don't think the extra logic or processing is worth the return. If anyone chooses to rely on undocumented props, it's at their own risk. When we say we would return it, it sort of becomes a contract - that we must adhere to. And we can add the filtering logic anytime, but I'm hoping that we're not forced to. 

We should fix the `onAction` to return the action data though.

mrchief: And regarding global event handler vs local - if both work exactly the same, passing the same stuff, then we don't need to keep or support both. My take would be 

 - either that global stays as is and the local is the one that bubbles up extra action data.
 - or make global bubble up action data and remove local `onAction` handler altogether (between global and local I think keeping the global makes more sense as it is easier for the caller).

Thoughts?

ellinge: A global event + passing of NodeAction should be sufficient to handle most use cases. Could be bloated if you have many different actions but I guess the use case is generic most of the times.

Then we don't have to worry about the handler/function and deep cloning at the moment.

I suggest we have the following implementation done instead, to be consistent:
```ts
onChange?: (currentNode: TreeNode, selectedNodes: TreeNode[]) => void;
onAction?: (currentAction: NodeAction, currentNode: TreeNode) => void;
onNodeToggle?: (currentNode: TreeNode) => void;
```

mrchief: Sold!

ellinge: We can create a separate issue then for that then or do you want it fixed here? I think it's quite easily solved without the local onaction.

mrchief: I was pondering the same. I think a separate one would be better (to keep the changeset small)

ellinge: Maybe merge this but not release it yet then? Instead of branching from this. Then the typings can be added for radioSelect also in the other PR.

mrchief: Yeah, let's merge this without releasing it and then proceed with the rest. Will do so as soon as I get a chance.

mrchief:
Many thanks for adding this!

PR #224, opened 2019-03-27
style: Auto format code using prettier before commit

mrchief:
## What does it do?

Adds prettierrc and pre commit hooks to format code automatically. Cleans up eslintrc and switched to react-app lint rules


package.json
```diff
@@ -26,13 +26,12 @@
     "build": "rimraf dist/**/* && cross-env NODE_ENV=production webpack --config webpack.config.js --bail --mode=production",
```
ellinge: Checked if we could remove any more dependencies with depcheck after changing linting. This is not 100% but some hints at least. As you can see some of these are false positives. Either way the tool reports:

Unused devDependencies
* @babel/plugin-syntax-jsx
* @babel/preset-stage-3
* @commitlint/cli
* @commitlint/config-conventional
* babel-plugin-transform-react-remove-prop-types
* conventional-changelog-cli
* eslint-config-prettier
* husky
* ignore-styles
* jsdom
* jsdom-global
* lint-staged
* postcss
* prettier-stylelint
* raw-loader
* style-loader
* stylelint-config-standard
* url-loader
* webpack-cli
* why-did-you-update

mrchief: > Checked if we could remove any more dependencies with depcheck after changing linting.

Glad I'm not the only dependency-averse person out there :)

I've given up on depcheck years ago. I did give it another try yesterday but nothing has changed. That said, I did check the list and I know where each one of those deps are used, so you're right about them being false positives.

mrchief:
@ellinge I'm going to hold on to this PR till we merge the older ones first. That way, we have to solve conflicts in one branch as opposed to solving in 3 otherwise.

mrchief:
Maybe we should merge this now? I don't think I can ever catch up to you :)

ellinge:
Sure! Hoped it was the radioSelect @ first from the mail notification... 🥈

mrchief:
> Sure! Hoped it was the radioSelect @ first from the mail notification... 🥈

Alright, alright! We'll do this after radio select then 😃 

> Seems to be some build errors though.

We need to pretty things after every merge 🎀

PR #225, opened 2019-03-27
feat: Add keyboard navigation

### LINKED ISSUE #216: WCAG support
There's currently no support for keyboard navigation and no ARIA-attributes present.
Here are some guidelines one should consider complying with WCAG:
https://www.w3.org/TR/wai-aria-practices/examples/listbox/listbox-collapsible.html

Here are some of the highlights on attributes one should consider implementing:
- role
- aria-haspopup
- aria-expanded
- aria-selected

ellinge:
Work in progress.

Adds keyboard navigation as an option. Inspired by:
https://www.w3.org/TR/wai-aria-practices-1.1/examples/treeview/treeview-1/treeview-1a.html
https://www.w3.org/TR/wai-aria-practices-1.1/examples/combobox/aria1.1pattern/listbox-combo.html
Also have looked at react-select

Also adds a11y-tests with axe. The only violation identified is missing label for seachInput though.
Added possibility to pass label with id which sets aria-labelledby or else aria-label on searchInput.

Added some related aria-attributes based on examples. Probably more needed though. This is mostly covered to handle traversing of nodes with keyboard and not how we should handle tags and actions by keyboard. This is something to start with at least.

Sample up and running here:
https://ellinge.github.io/react-dropdown-tree-select-test/#DevelopTemp-nocheckeddefault

Turned off can be seen here:
https://ellinge.github.io/react-dropdown-tree-select-test/#DevelopTemp-keyboardnavigationoff

Partly addresses #216 

- [x] New feature


README.md
```diff
@@ -52,6 +52,8 @@ A lightweight and fast control to render a select component that can display hie
   - [showDropdown](#showDropdown)
   - [form states (disabled|readOnly)](#formstates)
   - [id](#id)
+  - [label](#label)
+  - [enableKeyboardNavigation](#enableKeyboardNavigation)
```
mrchief: I think this should be on always.

README.md
```diff
@@ -331,6 +333,21 @@ Specific id for container. The container renders with a default id of `rdtsN` wh
...
+
+Type: `string`
+
+Adds `aria-labelledby` to search input when input starts with `#`, adds `aria-label` to search input when label has value (not containing '#')
+
+### enableKeyboardNavigation
+
+Type: `bool` (default: `false`)
+
+Adds navigation with `arrow` keys, `page down/up` / `home/end` and toggle of selection with `enter`. `Arrow/page up/down` also toggles open of dropdown if closed.
```
mrchief: Will it work uniformly across different OSes? In theory, it should but we know in practice things aren't always as simple. I can help test this on Windows, Mac but I'm wondering if we need to setup some integration tests (not sure if there are open source tools for doing this). Any way, food for thought.

README.md
```diff
@@ -331,6 +333,21 @@ Specific id for container. The container renders with a default id of `rdtsN` wh
...
+
+To close open dropdown `escape` or `tab` can be used and `backspace` can be used for deletion of tags on empty search input.
```
mrchief: `tab` for closing as in tabbing out? I'm Ok with just `esc` for closing which is what the native controls do.

package.json
```diff
@@ -55,12 +55,14 @@
...
     "all-contributors-cli": "^5.4.1",
     "ava": "1.0.0-beta.4",
+    "axe-core": "^3.2.2",
     "babel-core": "6.24.1",
     "babel-eslint": "8.0.0",
     "babel-loader": "7.1.2",
     "babel-plugin-istanbul": "4.1.3",
     "babel-plugin-transform-class-properties": "6.24.1",
     "babel-plugin-transform-react-remove-prop-types": "0.4.13",
+    "babel-polyfill": "^6.26.0",
```
mrchief: Do we need this? If yes, won't we need it for client-side as well? This will have a huge impact on bundle-size if we need to bundle it. We can list it as a peer dep or one of the things one needs to do to get keyboard navigation working but it'd be sweet if we can avoid requiring any polyfills.

src/tree-node/index.css
```diff
@@ -21,6 +21,10 @@
       }
     }
   }
+
+  &.focused {
```
mrchief: This is new and even if we wanted to add this, can we not use `:focus` selector?

package.json
ellinge: This is "only" a dev-dependency to run async await in tests.
Won't affect bundle size.

README.md
ellinge: Well if you've tabbed in you probably will tab out. But perhaps one should select the current "focused" item also then? But that should probably only happen for simpleSelect.

Tab into this example, use arrow down and tab out. It selects the current focused item.
https://www.w3.org/TR/wai-aria-practices/examples/listbox/listbox-collapsible.html

src/tree-node/index.css
ellinge: No because the focus is actually on the input still. 
See https://www.w3.org/TR/wai-aria-practices-1.1/examples/combobox/aria1.1pattern/listbox-combo.html

We use aria-activedescendent to specify current "focused" item.

README.md
ellinge: We use .key on the event. See https://www.w3.org/TR/uievents-key/#named-key-attribute-values
I hope this should be generic across OS and browsers.

There's one exception, to trigger toggle with space on <Toggle.
` if (e.key === 'Enter' || e.keyCode === 32)` where 32 is space. But this is actually only useful if keyboardnavigation is off since you can't tab there otherwise.

README.md
mrchief: For simple select - yes.

For multi, tabbing becomes tricky - if I focus on the input and tab, am I tabbing out or tabbing to next tag? Tabs cycle through controls so one could argue that it should go to next tag and then to its `x` button as well. But then, we can probably skip cycling to `x` if one can `tab > select tag > del > delete tag`. 

Maybe we should first write up a spec of what the keyboard navigation should be like in different cases - it'll help us validate our code as well.

Eventually, it can go in the Readme as keyboard map help section.

package.json
mrchief: Well, in that case, have you tried https://babeljs.io/docs/en/babel-plugin-transform-runtime/?

README.md
ellinge: We [accomplished this](https://github.com/dowjones/react-dropdown-tree-select/pull/233#discussion_r271608037) yesterday in #233 I think. It does not seem to be pushed yet though.
We added a keydownevent on the now "tabable" trigger-a. This and the click event now triggers handleClick as before. But we added a callback checking if the dropdown is visible after the state change and then sets focus on the input-field in this case. This makes it possible to tab into the dropdown. Clicking enter, space or arrowdown you get focus on the search input and the dropdown is visible. Tabbing takes you to the next delete-button if any. After the tag-deletes you get focus on the input if tabbing further.

package.json
ellinge: Tried now, got  Error: Requires Babel "^7.0.0-0", but was loaded with "6.26.3". If you are sure you have a compatible version of @babel/core, it is likely that something in your build process is loading the wrong version. Inspect the stack trace of this error to look for the first entry that doesn't mention "@babel/core" or "babel-core" to see what is calling Babel.

ellinge: Only seems compatible with 7 of babel:
https://www.npmjs.com/package/@babel/plugin-transform-runtime

Do you want me to upgrade babel-version? It set @ 6.24.1

ellinge: Switched to 
```json
["transform-runtime", { "polyfill": false, "regenerator": true }]
```

README.md
ellinge: Now it's in place you can see the combined result here. https://ellinge.github.io/react-dropdown-tree-select-test/#DevelopTemp-default

package.json
mrchief: Umm, yeah there's something for babel 6 as well. Glad you found it.

src/tree-manager/index.js
```diff
@@ -222,28 +219,43 @@ class TreeManager {
...
+    const prevFocus = currentFocus && this.getNodeById(currentFocus)
+    const action = keyboardNavigation.getAction(prevFocus, key)
 
-      if (node.checked) {
-        tags.push(node)
+    if (FocusActionNames.has(action)) {
+      const newFocus = this.handleFocusNavigationkey(tree, action, prevFocus, markSubTreeOnNonExpanded)
+      return newFocus
+    }
+    if (action === NavActions.ToggleChecked && prevFocus && !readOnly && !(prevFocus.readOnly || prevFocus.disabled) && tree.has(prevFocus._id)) {
```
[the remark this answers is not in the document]
mrchief: I agree with the bot here :)

src/tag/index.js
```diff
@@ -23,12 +23,16 @@ class Tag extends PureComponent {
...
+    const tagId = `${id}_tag`
+    const className = cx('tag-remove', { readOnly }, { disabled })
+    const onClick = !readOnly && !disabled ? this.handleClick : undefined
 
     return (
-      <span className={cx('tag')}>
+      <span className={cx('tag')} id={tagId}>
         {label}
-        <button onClick={!readOnly && !disabled ? this.handleClick : undefined} className={cx('tag-remove', { readOnly }, { disabled })} type="button">
+        <button onClick={onClick} className={className} type="button" aria-label="Delete" aria-labelledby={tagId} aria-readonly={readOnly || disabled}>
```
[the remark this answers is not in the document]
mrchief: Isn't `aria-readonly` meant for inputs, textarea, and content editable elements only?

README.md
mrchief: Nice! I see that tabbing between input and tags work beautifully. But neither backspace not delete had any effect on the tags. Tabbing also doesn't go to the `x` button but I think that's ok if backspace and/or delete can delete those. 

On the other hand, `tag > enter > delete` doesn't feel right. Enter is never associated with delete unless the focus is explicitly on a delete button (in our case the `x` button). Which is another reason I think `x` should not be tab-able - it just feels like too many tabs to navigate.

README.md
mrchief: Let me see if there are any tools out there that can help us set up some test suites. I don't think setting up test suites is a blocker for this PR though.

mrchief: Also, I can't seem to find anything - did we decide on keeping the flag `enableKeyboardNavigation`? I don't see any reason why we would need it - keyboard navigation should come naturally.

src/utils/nodeVisitor.js
```diff
@@ -0,0 +1,41 @@
+import isEmpty from './isEmpty'
+
```
mrchief: I like the idea of `nodeVisitor.js`. It belongs to `tree-manager` though.

src/utils/keyboardNavigation.js
```diff
@@ -0,0 +1,151 @@
+import nodeVisitor from './nodeVisitor'
+
```
mrchief: I like how you came up with this idea of keeping keyboard navigation to this one single place - this is a great design in terms of separation of concerns. I would move it out of utils though. In general, I'm not a fan of `utils` or `common` or `shared` folders in any project. Over time, they become a dumping ground of sorts (I know we have a bunch of stuff in utils and more keep creeping up, but trust me - if I could find a better place for them, I would 😃).

Since this deals with navigating the tree, I think it should be in `tree` or `tree-manager`. Given how little it has to do with any UI elements and how much it has to do with its "utility" nature, I think `tree-manager` is a better place between the two.

The same goes for `nodeVistor` and aria helpers should go to a new folder called `a11y`. In general, anything that is more descriptive than the vague `utils` is preferable.

Thoughts?

src/index.keyboardNav.test.js
```diff
@@ -0,0 +1,132 @@
...
+      ...node('c3', 'ccc 3'),
+      children: [node('a3', 'aaa 3'), node('b1', 'bbb 1'), node('b2', 'bbb 2')]
+    }
+  ]
+}
+
+test('some key presses opens dropdown on keyboardNavigation', t => {
+  ['ArrowUp', 'ArrowDown', 'Home', 'PageUp', 'End', 'PageDown', 'a', 'B'].forEach(key => {
+    const wrapper = mount(<DropdownTreeSelect data={tree} enableKeyboardNavigation />)
+    wrapper.find('.search').simulate('keyDown', { key })
```
mrchief: Since I see a lot of `simulate` being used in tests, here's a fair warning if you aren't aware of it already - `simulate` doesn't simulate actual keyboard or browser events. Its fake su=imulation - it only wires of `<eventName> to an `on<eventName>` method/function in your component.

So if you want to actually test browser DOM events, Enzyme is not your tool. 

That said, I, however, tend to agree with Enzyme team's philosophy here - we should not be testing browser's functionality anyway; rather ensure that out handler does it job when called upon. And that can be done by simply calling `wrapper.instance().onXXX()` directly and asserting the expected state.

ellinge: I am aware. But now I test both that the handler is triggered and it that it does it's job. So not sure what difference it would make.

src/tree-manager/index.js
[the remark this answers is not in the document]
ellinge: So do I, but had other things to do 😄

src/utils/keyboardNavigation.js
ellinge: Agree with keyboardNavigation and nodeVisitor but hesistant about a11y. Currently only has "one helper" and things needed for axe-testing.

README.md
ellinge: It does go to the x-button but I've highligheted the current tab also. So not sure what to make of your last paragraph.

README.md
ellinge: I can remove it

src/utils/keyboardNavigation.js
mrchief: I hear ya and I'll let you be the judge here. All I'm saying is to consider not having more stuff in utils ***if*** possible.

src/index.keyboardNav.test.js
mrchief: > So not sure what difference it would make.

The difference is that we're not depending on a flaky API (`enzyme.simulate`). 

> I've used cypress and testcafe for some uitesting but I hardly think this issue is the one to introduce that. Too many things to cover with ci, what browsers etc.

With you 100%. That was more of a phase 2 thingy where we set up some real-world tests that mimic an actual user. I imagine those tests should outside of any components and should not be aware of the code at all. My thinking is that as we're introducing more features, we need more checks & balances to ensure stable releases.

README.md
ellinge: I added keydown for backspace and delete now also when focusing on a tag btw.
The test page is updated. See if it stills feels wrong?

ellinge: Moved out keyboard nav to a separate section in the readme also. Not sure how detailed it should be and still be clear.

mrchief: > Not sure how detailed it should be and still be clear.

We won't need a detailed section if the navigation is natural and intuitive :smile:

Some observations (and by the way - I love your example page - makes it so easy to test out the changes):

 - If I hit space here, it goes into search mode. We're toggling the checkboxes with but usually, space toggles them for native controls also. Is this a focus issue? I think if the label or checkbox has the focus, then space would work but for some reason, input has the focus.
image

 - `tab > tag > delete > focus gone`. One needs to tab again to select the next one or enter the search box. Will it be too complex to move the focus forward - to whatever is next (select next tag if present else move the focus to search box)? 

 - There was also another case which I can't seem to reproduce now - what happend was I had the dropdown open, and it stayed open somehow when I moved to next one. I shift tabbed to move to previous and it first went to firt dropdown item and then shift tab moved to the search box and so on. So shift tabbing worked but it felt odd (even though it was the right order). I expected it to tab to previous control (and not a dropdown item). I can't reproduce it now though.

mrchief: Speaking of Space key toggling checkbox, are we currently handling the `partialSelect` use case? A parent level checkbox can have a partial selected state and space will need to cycle through three states. Will need `aria-checked="mixed"` attributes as well during partial state.

ellinge: Well space is also part of the possible search entry. So normally yes. You would select with space. But not when you have a input query. See:
https://www.w3.org/TR/wai-aria-practices-1.1/examples/combobox/aria1.1pattern/listbox-combo.html

src/utils/keyboardNavigation.js
ellinge: a11y moved

src/index.keyboardNav.test.js
ellinge: Unfortunely could not get the same result without simulate even if all the needed props should be passed in the event. I think it's a state-race (sending new keys too soon). Don't want to modify the code to have a callback etc just because I need the test to work.

I've commented out the tries if you want to give it a go.

README.md
ellinge: Added focus after delete-logic now. Added it in onTagRemove and had to use getElementById to find the next target. Not sure how we would pass refs etc up but that would be an option as well.

ellinge: I'll look into handling tabbing better also. For simpleSelect. Toggle check (only select, not deselect) on tab.
For the other we either do it like now, just leave, or we handle it like next prev node. One option would be to tab to any action in this case also.

I also need to fix shift-tab to go to previous tag if any. Just not sure how we should handle it if tabbing also navigates tree. Escape and shift-tab? Would be simpler if we just leaved.

src/tag/index.js
```diff
@@ -41,14 +41,15 @@ class Tag extends PureComponent {
...
+    const isDisabled = readOnly || disabled
+    const onClick = !isDisabled ? this.handleClick : undefined
+    const onKeyDown = !isDisabled ? this.onKeyDown : undefined
+    const onKeyUp = !isDisabled ? this.onKeyUp : undefined
 
     return (
       <span className={cx('tag')} id={tagId}>
         {label}
-        <button onClick={onClick} onKeyDown={onKeyDown} onKeyUp={onKeyUp} className={className} type="button" aria-label="Delete" aria-labelledby={tagId} aria-disabled={readOnly || disabled}>
+        <button onClick={onClick} onKeyDown={onKeyDown} onKeyUp={onKeyUp} className={className} type="button" aria-label="Delete" aria-labelledby={tagId} aria-disabled={isDisabled}>
```
mrchief: Do we need both `aria-label` and `aria-labelledby`?

ellinge: When I inspected it we get the "{Tag} Delete" as description.
Otherwise it would be unclear what the button does.

mrchief: > I guess we only could set a aria-label="Delete {Label}" instead?

Sure, let's roll with that. Umm, do think we should say `Delete {Label} tag`?

README.md
ellinge: I've only fixed so tab also selects current focused node on simpleSelect now. Had some issues in firefox with clicking input field and shift-tab. Did not go to lasts tag.

src/tag/index.js
ellinge: Should be able to set by prop also I guess. But should that be in the format "Delete {Label} tag" in that case?

src/index.js
```diff
@@ -197,25 +197,31 @@ class DropdownTreeSelect extends Component {
...
       }
     } else if (showDropdown && ['Escape', 'Tab'].indexOf(e.key) > -1) {
-      // Triggers close
-      this.keepDropdownActive = false
-      this.handleClick()
+      if (simpleSelect && tree.has(currentFocus)) {
+        this.onCheckboxChange(currentFocus, true)
+      } else {
+        // Triggers close
+        setTimeout(() => {
```
mrchief: `setTimeout`? I haven't looked at the code too deeply but marking this to revisit later.

ellinge: Had issues in firefox with clicking a seach input and shift-tab did not focus the last tag. Worked in chrome but required setTimeout with firefox. I'll have another look later.

src/tag/index.js
ellinge: bild
From the pill list here, I'll guess we'll keep it then. But perhaps one should be able to change it by prop (localization or other reasons).

ellinge: bild
From the pill list [here](https://www.w3.org/TR/2017/NOTE-wai-aria-practices-1.1-20171214/examples/grid/LayoutGrids.html), I'll guess we'll keep it then. But perhaps one should be able to change it by prop (localization or other reasons).

ellinge:
https://media.giphy.com/media/3orifdJUNXBWPLhNxS/giphy.gif
Almost got codeclimate running locally but not all the way unfortunately. Planned to set filters for exclusion on propTypes-assignments and similar.

ellinge:
Celebrate good times
Almost got codeclimate running locally but not all the way unfortunately.
Planned to set filters for exclusion on propTypes-assignments and similar.

src/tag/index.js
mrchief: Hmmm, I'm not sure why both are needed. Will look up the reasons later. For now, let's keep them both.

mrchief: > But should that be in the format "Delete {Label} tag" in that case?

Think so. I was imagining how the screen reader would read it out and so, adding "tag" would help or not.

ellinge: I suggest we go with the matching example from w3 (which uses Remove + two label references by id) .Added possibility to control the label by prop though.

mrchief: Yeah, let's do that. 

> We could default it to "Remove tag"?

Default which one?

ellinge: `labelRemove = "Remove"`	> "Remove tag"

src/index.keyboardNav.test.js
ellinge: Did you have a go on this? Should remove the commented code either way.

src/tag/index.js
mrchief: Oh, I see. I missed those new props somehow. 

I don't think it `labelRemove` should be an option. Are you thinking this in terms of i18n? 

And can you tell me more about `label` prop? Why is it needed?

ellinge: For i18n and validation. It's in the readme/typings. When the trigger has a role it also will require a label just as the input.

src/index.js
ellinge: ### Focus click and shift-tab behaviour
✔️ **Chrome no timout**

chromeNoTimeout

⛔️  **Firefox no timout**

ffNoTimeout

✔️ **Firefox with timeout**

ffSetTimeout

src/tag/index.js
mrchief: What do you say we tackle i18n separately? And if possible in a way that doesn't require more props. I already think that we're approaching a "prop overload" (I'm wondering if it's time to split the component into separate packages) and so if we go the prop way, I fear we may have to introduce more props later.

ellinge: How about starting to bundle props in objects instead, Mode/enum instad of separate booleans, `texts = { placeholder, label, labelRemove }` etc?

We would need a possibility to set the label since "Remove Blåval" and "Remove Bläckfisk" is not something we want presented for our potential screen reader users. Having to modify the dom because the component does not provide a way feels a bit odd. It's the only "static" text then which you would not be able to change without the prop.

mrchief: > How about starting to bundle props in objects instead, Mode/enum instad of separate booleans, texts = { placeholder, label, labelRemove } etc?

That seems like a good start. Let's go with that for now. We won't be backward compatible but I think that's Ok.

ellinge: I'll try to fix the migration step guide in the onAction pr so we can get that through and fill it with prop-changes then.

src/index.keyboardNav.test.js
mrchief: For some reason, I didn't even see your earlier comment. I think Github only bring you to the latest comment. If it works with `simulate` then let's go with it.

src/index.js
mrchief: Thanks for the GIFs. Am I not able to focus properly or both the cases with FF look the same?

src/a11y/index.js
```diff
@@ -0,0 +1 @@
+export { getAriaLabel } from './a11y.utils'
```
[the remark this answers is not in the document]
mrchief: Ignore these... CC doesn't have these errors on their site - this comment is result of bad timing or something.

src/index.js
ellinge: I focus with a mouse click and then shift-tab. Chrome and ff with setTimeout focuses on the tag. Ff without focuses on the hierarchial-checkbox.

src/tag/index.js
ellinge: The bundle of props is now reverted according [the other PR comment](https://github.com/dowjones/react-dropdown-tree-select/pull/217#issuecomment-483472154)

mrchief:
@ellinge, help me figure out the merge sequence, please! Ideally, it shouldn't matter, but since we have overlapping PRs, I'm trying to minimize the effort here. What do you say to this order:

 1. 225, followed by
 2. 233 - do a release
 3. 230, followed by 
 4. 242 - do a release

ellinge:
@mrchief sounds good. Will probably need more tweaking (WCAG) but perhaps it better to start off with something than waiting forever for the perfect PR. Haven’t really had the time or energy to do any more tries with actions and the tag-navigation.

The good thing is that even if we need to change something with aria or keyboard no one should really rely on them (css selectors or from outside in code I mean).

mrchief:
> @mrchief sounds good. Will probably need more tweaking (WCAG) but perhaps it’s better to start off with something than waiting forever for the perfect PR. Haven’t really had the time or energy to do any more tries with actions and the tag-navigation.

> The good thing is that even if we need to change something with aria or keyboard no one should really rely on them (css selectors or from outside in code I mean).

Yep & Yep! So I'm gonna start in that order and try to close the PRs.

src/index.js
```diff
@@ -193,6 +210,54 @@ class DropdownTreeSelect extends Component {
...
+  onKeyboardKeyDown = e => {
+    const { readOnly, simpleSelect } = this.props
+    const { showDropdown, tags, searchModeOn, currentFocus } = this.state
+    const tm = this.treeManager
+    const tree = searchModeOn ? tm.matchTree : tm.tree
+
+    if (!showDropdown && (keyboardNavigation.isValidKey(e.key, false) || /^\w$/i.test(e.key))) {
+      // Triggers open of dropdown and retriggers event
+      e.persist()
+      this.handleClick(null, () => this.onKeyboardKeyDown(e))
```
mrchief: Help me understand why we're doing it this way - especially why we're calling `onKeyboardKeyDown` again within the `onKeyboardKeyDown` handler.

I believe you're trying to reuse the state setting logic and then react after that logic does its thing?  If so, I think it's time to refactor `handleClick`.

src/index.js
```diff
@@ -193,6 +210,54 @@ class DropdownTreeSelect extends Component {
...
+        this.onCheckboxChange(currentFocus, true)
+      } else {
+        // Triggers close
+        this.keepDropdownActive = false
+        this.handleClick()
+      }
+      return
+    } else if (
+      e.key === 'Backspace' &&
+      tags &&
```
mrchief: Is it possible to get rid of these (seemingly) excessive branches? 

Something like:
```
      e.key === 'Backspace' &&
      tags.length &&
      this.searchInput.value.length === 0
```

`tags` can be eliminated if we set it to the empty array in the constructor - not sure if that would have other ramifications and by the time this handler executes, `this.searchInput` should be valid.

src/index.keyboardNav.test.js
```diff
@@ -0,0 +1,163 @@
...
+  const wrapper = mount(<DropdownTreeSelect data={tree} showDropdown />)
+  t.false(wrapper.find('li.focused').exists())
+  triggerOnKeyboardKeyDown(wrapper, ['ArrowDown'])
+  t.deepEqual(wrapper.find('li.focused').text(), 'ccc 1')
+  wrapper.setProps({ data: tree })
+  wrapper.instance().handleClick()
+  t.deepEqual(wrapper.find('li.focused').text(), 'ccc 1')
+})
+
+test('should set current focus as selected on tab out for simpleSelect', t => {
```
mrchief: Kudos for adding these tests! 🙏

src/index.js
```diff
@@ -139,8 +149,13 @@ class DropdownTreeSelect extends Component {
     })
   }
 
-  onTagRemove = id => {
-    this.onCheckboxChange(id, false)
+  onTagRemove = (id, isKeyboardEvent) => {
```
mrchief: I don't see any call site that passes `isKeyboardEvent`. What am I missing?

src/index.js
```diff
@@ -193,6 +210,54 @@ class DropdownTreeSelect extends Component {
...
+    const { readOnly, simpleSelect } = this.props
+    const { showDropdown, tags, searchModeOn, currentFocus } = this.state
+    const tm = this.treeManager
+    const tree = searchModeOn ? tm.matchTree : tm.tree
+
+    if (!showDropdown && (keyboardNavigation.isValidKey(e.key, false) || /^\w$/i.test(e.key))) {
+      // Triggers open of dropdown and retriggers event
+      e.persist()
+      this.handleClick(null, () => this.onKeyboardKeyDown(e))
+      if (/\w/i.test(e.key)) return
```
mrchief: `/\w/i.test(e.key)` is not Unicode safe. [This](https://mathiasbynens.be/notes/es6-unicode-regex#impact-character-class-escapes) makes my head spin, especially the note at the bottom. 

What is the intention here? If possible, we should look for non-regex based solutions.

src/index.js
```diff
@@ -177,7 +192,9 @@ class DropdownTreeSelect extends Component {
       document.removeEventListener('click', this.handleOutsideClick, false)
     }
 
-    this.setState(nextState)
+    this.setState(nextState, () => {
+      callback && callback(tags)
```
mrchief: This would need `typeof callback === 'function' && callback(tags)` if you truly wanna be defensive but in general, I think we should see if we can eliminate the need to rely on `setState` callbacks. We'll probably have to refactor existing event handlers in a way that the setState portion can be extracted out into a reusable block - but I'm not sure if the resulting code would be better than having the callbacks. I'll take a deeper dive tomorrow but just leaving this note out here in case you have any ideas about it.

src/index.js
ellinge: I wait for the state to update and retrigger the event to select the first node in the now visible dropdown.

src/index.js
ellinge: We return to avoid skipping (preventDefault) the event if it's a char-press.
Probably can be solved in other ways.

src/index.js
ellinge: Nope, no other ideas, have to wait for state to be set. This is the fastest way there.

src/index.js
ellinge: https://github.com/dowjones/react-dropdown-tree-select/blob/5edbc88ddb8c39e8966b3b04cb8832a210297cb7/src/tag/index.js#L23

ellinge: If you delete by keyboard we set the next focus, not if clicked by mouse.

src/index.js
ellinge: getTags seems safe to always return an array so should be fine.

mrchief:
Looks like `babel-runtime` is getting bundled which caused the bundle size to jump to 19KB

image

I'm checking if we can get rid of it.

PR #230, opened 2019-03-31
fix: Updated onAction to bubble up custom props

ellinge:
## What does it do?

When adding typings in #223 it was found that the current implementation of onAction is not working.
Because of "deepcloning" the local onAction-events does not work anymore as intended.

The global event worked though but was somewhat limited. The only thing passed was the generated actionid (if not set as prop) and nodeid. See:
bild

To be consistent with the other events the global action is now changed to pass the action-object and current node instead, or defined with TypeScript instead:
```ts
onChange?: (currentNode: TreeNode, selectedNodes: TreeNode[]) => void;
onAction?: (currentAction: NodeAction, currentNode: TreeNode) => void;
onNodeToggle?: (currentNode: TreeNode) => void;
```
Where NodeAction and TreeNode is:
https://github.com/dowjones/react-dropdown-tree-select/blob/051d5ed34e61cc0098b2946396b222aa73d925ad/types/react-dropdown-tree-select.d.ts#L125
https://github.com/dowjones/react-dropdown-tree-select/blob/051d5ed34e61cc0098b2946396b222aa73d925ad/types/react-dropdown-tree-select.d.ts#L90

## Type of change

- [x] Breaking change


ellinge:
Up and running under events here:

mrchief:
Looks good! 

Since it's a breaking change, I guess we need to update the title to reflect that so that versions are bumped accordingly.

Also, should we provide a migration guide about new action syntax? It's not much but usually is expected for a major version bump.

mrchief:
> Since it's currently not working on local and you only get passed the action/node id 

It'd be working for those on v1.0.4 or older. Hence the dilemma.

> Where should this migration step be added?

How about `docs/guides/migrating-from-1-to-2.md`? And link it via the PR's description (not sure about this part doing a dry-run with semantic-release will tell us more).

ellinge:
> 
> 
> > Since it's currently not working on local and you only get passed the action/node id
> 
> It'd be working for those on v1.0.4 or older. Hence the dilemma.
> 
> > Where should this migration step be added?
> 
> How about `docs/guides/migrating-from-1-to-2.md`? And link it via the PR's description (not sure about this part but doing a dry-run with semantic-release will tell us more).

Tried v1.04 here but can't get it working onLocal either. I can see that the onGlobal-properties is different from v1.17 as well.
https://ellinge.github.io/react-dropdown-tree-select-test/#v104-onaction

Perhaps it's better you write the migration step since you have the history of it's previous workings?

mrchief:
> Tried v1.04 here but can't get it working onLocal either. I can see that the onGlobal-properties is different from v1.17 as well.
ellinge.github.io/react-dropdown-tree-select-test/#v104-onaction

Thanks for taking the effort. It could be buggy from beginning. We can check but I think it'll be a fruitless exercise. 

> Perhaps it's better you write the migration step since you have the history of it's previous workings?

Your description is pretty accurate but I'll let you know if I find anything.

ellinge:
I've added a migration doc now.

docs/guides/migrating-from-1-to-2.md
```diff
@@ -0,0 +1,38 @@
+# React Dropdown Tree Select Migration Guide v1 to v2
+
+We've made some changes in the component between the versions which is covered in the following guide
+
+## Property changes
+
+| Description                                                              | Usage before                                                | Usage after                                                   |
+| ------------------------------------------------------------------------ | ----------------------------------------------------------- | ------------------------------------------------------------- |
+| Added a new radio select mode<br>Selection is done in a shared mode-prop | `simpleSelect={true}` / `simpleSelect`                      | `mode='simpleSelect'`                                         |
```
mrchief: We can release radio select before the other two. That way, people can still get the most out of v1 without upgrading to v2. radio select is backward compatible which is another reason I prefer it to be part of a v1 release. Thoughts?

ellinge: See [other pr:s comment](https://github.com/dowjones/react-dropdown-tree-select/pull/217#issuecomment-483514803). The same applies here.

ellinge: And now it's reverted

mrchief:
@ellinge Made some tweaks - primarily to make `onAction` signature consistent with remaining handlers. 

Also, I think there is a way to include the migration steps in the release log itself. Needs some experimentation to be sure.

mrchief:
Confirmed. We don't need a separate guide. I can add the breaking change description in the release log itself.

PR #233, opened 2019-04-02
feat: Add ARIA roles and attributes

### LINKED ISSUE #216: WCAG support

gazab:
## What does it do?

Adds some ARIA roles and attributes according to: 
- https://www.w3.org/TR/wai-aria-practices/#listbox_roles_states_props
- https://www.w3.org/TR/wai-aria-practices/examples/listbox/listbox-collapsible.html

Does not add label to input element because #225 already solves that.

## Fixes # (issue)

Partly addresses #216 

## Type of change

- [ ] New feature (non-breaking change which adds functionality)


__snapshots__/src/index.test.js.md
```diff
@@ -16,8 +16,12 @@ Generated by [AVA](https://ava.li).
         className="dropdown"
       >
         <a
+          'aria-expanded'={false}
+          "aria-haspopup"="listbox"
```
mrchief: Interesting how the renderer emits single quotes for a pro and double quotes for another.

src/tree-node/index.test.js
```diff
@@ -94,3 +94,31 @@ test('should render data attributes', t => {
...
+    _id: '0-0-0',
+    _parent: '0-0',
+    label: 'item0-0-0',
+    value: 'value0-0-0',
+    className: 'cn0-0-0',
+    checked: true
+  }
+
+  const wrapper = mount(<TreeNode {...node} />)
+  t.is(wrapper.prop('aria-selected', true))
```
mrchief: Can we generate a snapshot for this instead of checking specific props? Nothing wrong with this in general, but just to be consistent with other tests in here.

mrchief:
This is great! Thanks for sending this!

src/index.js
```diff
@@ -201,7 +201,7 @@ class DropdownTreeSelect extends Component {
         }}
       >
         <div className="dropdown">
-          <a className={dropdownTriggerClassname} onClick={!this.props.disabled && this.handleClick}>
+          <a className={dropdownTriggerClassname} onClick={!this.props.disabled && this.handleClick} role="button" tabIndex="0" aria-haspopup="listbox" aria-expanded={this.state.showDropdown}>
```
ellinge: Noticed that you cannot trigger the click with keyboard (Enter, Space) when you're now able to tab to the control.
We probably should add a onKeyDown which checks if Enter or Space is entered which focuses the input field instead.
The default action onclick should be focus on the input imho also.

Then you show the dropdown as before but also can start typing right away to filter the results. When #225 is in place you also can navigate to nodes with keyboard.

src/tree-node/index.js
```diff
@@ -88,7 +88,7 @@ class TreeNode extends PureComponent {
     const style = keepTreeOnSearch || !searchModeOn ? { paddingLeft: `${(_depth || 0) * 20}px` } : {}
 
     return (
-      <li className={liCx} style={style} {...getDataset(dataset)}>
+      <li className={liCx} style={style} {...getDataset(dataset)} role="option" aria-selected={checked}>
```
ellinge: Perhaps we should add a function getAriaTags, {...getAriaTags(props)}. It would make it easier to check on simpleSelect since I think we need different roles depending on type of dropdown. `role="option"` is fine when emulating a regular dropdown, but when we have an expandable tree we probably need to use `role="row"` and `role="tree"` on the "parent" instead.

Then we can add `aria-level={_depth +1}` and` aria-expanded={_children && expanded}` for this also which is not valid with role="option".

We probably also should add `aria-readonly={readonly || disabled}`

src/tree/index.js
```diff
@@ -99,10 +99,10 @@ class Tree extends Component {
   }
 
   render() {
-    const { searchModeOn } = this.props
+    const { readOnly, searchModeOn } = this.props
 
     return (
-      <ul className={`root ${searchModeOn ? 'searchModeOn' : ''}`} ref={this.setNodeRef}>
+      <ul className={`root ${searchModeOn ? 'searchModeOn' : ''}`} ref={this.setNodeRef} role="listbox" aria-readonly={readOnly} aria-multiselectable={!this.props.simpleSelect}>
```
ellinge: See [previous comment for child/li](https://github.com/dowjones/react-dropdown-tree-select/pull/233/files?file-filters%5B%5D=.js#r271610907)
I think we probably need to use `role="treeitem"` and `role="tree"` on the "parent" instead if !simpleSelect.

We probably should pass disabled down also to when setting aria-readonly?

__snapshots__/src/index.test.js.md
gazab: Yeah, I've had some problems with snapshots being different depending on what machine I used. My Mac liked to output the markdown header with slash instead of backslash :/ Not sure what I can do about this

src/tree-node/index.test.js
gazab: Done! I had some problems with the tree test though so I removed it. I see both the other snapshots test there are skipped so they might all be problematic?

src/tree-node/index.js
gazab: Agreed and fixed!

src/tree-node/index.js
```diff
@@ -61,6 +61,20 @@ class TreeNode extends PureComponent {
...
 
+  getAriaAttributes = () => {
+    const { _children, _depth, checked, disabled, expanded, readOnly, simpleSelect } = this.props
+    const attributes = {}
+
+    attributes.role = simpleSelect ? 'option' : 'treeitem'
+    attributes['aria-selected'] = checked
+    attributes['aria-level'] = _depth || 0 + 1
+    attributes['aria-expanded'] = _children && expanded
+    if (simpleSelect) {
```
ellinge: A tree dropdown can have disabled/readonly nodes as well. Instead this should be negated around level and expanded which is not applicable on a flat dropdown.

src/tree-node/index.test.js
mrchief: The skipped ones have a pending issue open with ava - they cause the tests to hang for some unknown reason. 

When your changes result in updated snapshots, the tests will fail. You then need to verify whether the changes in the rendered HTML are expected or unintentional. If expected, you can update the snapshots by running `yarn test --update-snapshots` (see [more](https://github.com/avajs/ava/blob/master/docs/04-snapshot-testing.md)) and commit and push those snapshots. 

I see that the CI build has some failed snapshots tests. Those would pass once you checkin updated snapshots. Let me know if you have more questions.

mrchief:
Screen Shot 2019-04-05 at 4 45 09 PM

`aria-hidden` seems incorrect here as this is not hidden content. Also, I think we should add `aria-expanded` here as it marks collapsible content.

mrchief:
Screen Shot 2019-04-05 at 4 50 56 PM

The root should be level 0, instead it's level 1

ellinge:
> `aria-hidden` seems incorrect here as this is not hidden content. Also, I think we should add `aria-expanded` here as it marks collapsible content.
That's part of the other PR actually. We should however set tabIndex to -1 instead since you have expanded / arrow-toggleing on the treeitem instead. So that means it's hidden for screen readers.

Level is not zero-indexed, see:
https://www.w3.org/TR/wai-aria-practices-1.1/examples/treeview/treeview-2/treeview-2b.html

mrchief:
> So that means it's hidden for screen readers.

Yeah, I understood why you did it but then, why should it be hidden from screen readers?

> Level is not zero-indexed, see:

Good point. So then it should be level 1, and children should be +1 accordingly.

ellinge:
Your free to do changes you see fit. Other PR though.

mrchief:
> Reset it back to previous state perhaps? Then screen readers won’t find it either. 

Not sure I follow...

ellinge:
> Not sure I follow...

The current node is a non focusable i-tag containing no useful information what I can see for a screen reader, a.k.a ”hidden”.

mrchief:
> The current node (develop) is a non focusable i-tag containing no useful information what I can see for a screen reader, a.k.a ”hidden”.

Oh, I understand now. Thanks! So, apart from the level anomaly, we're otherwise good on this PR?

mrchief:
Don't bother resolving the intermediate conflicts. We can do a final resolving when this PR is about to be merged. Same goes for 230, 242

src/a11y/index.js
```diff
@@ -1,7 +1,7 @@
 export function getAriaLabel(label) {
   if (!label) return undefined
 
-  if (label && label.length && label[0] === '#') {
+  if (label && label[0] === '#') {
     return { 'aria-labelledby': label.replace(/#/g, '') }
```
mrchief: Might be useful to leave a comment here explaining why we're replacing the hash...

mrchief:
Need to figure out a way to resolve the codeclimate complexity error. I think CC is right here. Gonna take a shot with fresh mind tomorrow. Feel free to propose any ideas you may have.

src/a11y/index.js
ellinge: Check if that's good enough

ellinge:
Separated out the trigger to a separate js. Two large methods which could be moved as well.

src/index.js
```diff
@@ -259,19 +262,13 @@ class DropdownTreeSelect extends Component {
...
-      arrow: true,
-      disabled,
-      readOnly,
-      top: showDropdown,
-      bottom: !showDropdown,
-    })
 
     const activeDescendant = this.state.currentFocus ? `${this.state.currentFocus}_li` : undefined
 
+    const commonProps = { disabled, readOnly, activeDescendant, label, simpleSelect, radioSelect }
```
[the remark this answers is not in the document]
ellinge: Raised threshold. Until we find a way to ignore some of these object deconstructions we will probably have these kind of "false positives".

src/a11y/index.js
mrchief: > // If input label reference ids (hash), use a labelledby without hashes

Yes, but why? Why do we need to replace the hash - that's the question.

src/index.js
```diff
@@ -243,13 +252,7 @@ class DropdownTreeSelect extends Component {
...
-      e.key === 'Backspace' &&
-      tags &&
-      tags.length &&
-      this.searchInput &&
-      this.searchInput.value.length === 0
-    ) {
+    } else if (e.key === 'Backspace' && tags.length && this.searchInput.value.length === 0) {
       const lastTag = tags.pop()
       this.onCheckboxChange(lastTag._id, false)
     } else {
```
mrchief: What does this empty return accomplish? What'll happen if we removed it?

src/a11y/index.js
ellinge: Because of you write "hello world" you are providing the actual label as a text.
If you write "#hello #world" as a label you reference two label by id, hello and world. The aria-labelledby reference these without hash.

src/index.js
ellinge: It will avoid preventDefault to be triggered in cases where it did not match any if statement.

src/a11y/index.js
ellinge: Do you want a refence to https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-labelledby_attribute#Examples or similar?

mrchief: > It's explained in the documentation for the label-prop.

Exactly! The "what" is obvious while reading the code but the "why" is not. A comment that points out the why (it can be as simple as pointing out the readme#label prop) helps answer the "why".

I generally prefer the code to be self-documenting but sometimes, things are not obvious or clear just by reading the code. This is one such case IMHO. To the next person, "why is this hash replacement happening" is a question that is not obvious; which is why a comment is mighty helpful!

Makes sense?

mrchief:
@allcontributors[bot] Let's add @gazab for code

Issue #239, opened 2019-04-18
disable dropdown onClick search box


moonjy1993:
Hello,

Is there way to force dropdown content open? or like any disable when onClick of search box so that I can force the dropdown always expanded? if not, do you have any plan to implement such props? 

Thank you very much for reading this.

mrchief:
You can probably use [`showDropdown`](https://github.com/dowjones/react-dropdown-tree-select#showdropdown) to accomplish this.

moonjy1993:
unfortunately, `showDropdown` is only to intialiaze the state of dropdown. 
I am looking for a props that always keeps dropdown expanded.

mrchief:
@moonjy1993 I was thinking more like using `showDropdown` and hiding the dropdown toggle with CSS. But even with that, any outside click will close it. 

So, unfortunately, the answer to your question is no. 

Implementing this change seems [simple](https://github.com/dowjones/react-dropdown-tree-select/blob/develop/src/index.js#L102) though. Something like this should do the trick, I think:

```
const showDropdown = this.props.showDropdownAlways || this.keepDropdownActive || !prevState.showDropdown
```

I can get to it sometime but if you want this soon, feel free to send me a PR.

moonjy1993:
@mrchief  Hi, thank you for your response!
I will create a PR soon.

mrchief:
@moonjy1993 Great! Will look forward to it.

Got few things in my mind as I think about this:

 - would it be better to disable register the outsideClick handler for showDropdownAlways? 
 - Maybe we can bundle this new prop and `showDropdown` into a single one (not in this PR but the bundle props one - `showDropdown: initial|always`)
 - Or we can accept mixed values - true/false for current behavior and "always" for new behavior. This feels like a hack though and I'm not convinced if this is a good pattern (even though JavaScript will let us get away with it).

Thoughts? 

/cc @ellinge @Gregcop1

moonjy1993:
@mrchief 

- would it be better to disable register the outsideClick handler for showDropdownAlways?
 Yes. I guess that makes. I will also include that logic in PR.
   
- Maybe we can bundle this new prop and showDropdown into a single one (not in this PR but the bundle props one - showDropdown: initial|always)

- Or we can accept mixed values - true/false for current behavior and "always" for new behavior. This feels like a hack though and I'm not convinced if this is a good pattern (even though JavaScript will let us get away with it).

[...]

PR #241, opened 2019-04-18
feat: Ability to keep the dropdown open always (#241) ✨

moonjy1993:
## What does it do?

This adds a new prop: showDropdownAlways, and this disables the toggling of showDropdown when the user clicks search box.

## Fixes # (issue)

This closes issue #239. 

## Type of change

# Checklist:

- [x] My code follows the style guidelines of this project
- [ ] Added myself to contributors table
- [ ] I have updated `docs/bundle.js` file (ran `yarn build:docs` and checked in the new `bundle.js` file)


mrchief:
@moonjy1993 Thanks for submitting this. There are a few things we need to do before this PR can be merged:

 - Add unit tests for newly added functionality
 - Update Readme for the new prop
 - Update the PR description according to the checklist.

ellinge:
Probably should think about bundling this as well.
Change showDropdown to ’onfocus’ (default) | ’initial’ | ’always’ or something similar instead?

moonjy1993:
@ellinge 
but what happens if the user who's already using showDropdown to true in their code?

mrchief:
> Probably should think about bundling this as well.
> Change showDropdown to ’onfocus’ (default) | ’initial’ | ’always’ or something similar instead?

Yes, but we can make that part of #242

> but what happens if the user who's already using showDropdown to true in their code?

They would need a code update when they migrate to the new version, which is why it's a separate release.

moonjy1993:
@mrchief  okay. so I have updated the index.test.js and readMe files. will include showDropdownAlways for now, since you said the logic building will happen in other PR.

ellinge:
The typescript typings (.d.ts) needs an update as well. It’s a very newly added addition so there’s no checklist or similar which ensures this yet.

moonjy1993:
@ellinge  okay. just modified d.ts file.

mrchief:
> The typescript typings (.d.ts) needs an update as well. It’s a very newly added addition so there’s no checklist or similar which ensures this yet.

Gotta remember this myself! :)

src/index.test.js
```diff
@@ -72,6 +72,19 @@ test('shows dropdown', t => {
   t.snapshot(toJson(wrapper))
 })
 
+test('always shows dropdown', t => {
+  const { tree } = t.context
+  const wrapper = shallow(<DropdownTreeSelect id={dropdownId} data={tree} showDropdownAlways />)
+  t.snapshot(toJson(wrapper))
+})
+
+test('keeps dropdown open  for showDropdownAlways', t => {
```
mrchief: The extra space in the title can be removed.

mrchief: Other than that, this looks OK. @ellinge thoughts?

moonjy1993: @mrchief  hi, which line exactly are you referring to when you said "The extra space in the title"?

moonjy1993: yap. deleted the extra space.

ellinge: Probably could avoid some event handling etc if's always open, but I guess this is simple enough so never mind.

We have to make sure it's not closed with keyboard in some way also in the other PR, I think that it won't be a problem though. Not sure if we need to make any changes with ARIA as well but that's not part of this at this moment at least.

mrchief:
@allcontributors[bot] Let's add @moonjy1993 for code, bug

PR #242, opened 2019-04-18
feat: Group logically related props together (#242) ⚡️

ellinge:
## What does it do?

Bundles text props and "mode" (radio/simple) into a single object/prop

- [x] Breaking change


mrchief:
Can you base it off a branch in the repo itself? Or else update your fork for the latest prlint.json (the former is preferable though)

mrchief:
Ok, seems like it's not needed. Updating the branch did the trick. That makes me wonder if we should change that check (to be based off develop always)... Food for thought.

ellinge:
@mrchief nothing else we should bundle? showDropdown?

ellinge:
Also the doc changes is not reflected in the d.ts/typings

mrchief:
> @mrchief nothing else we should bundle? showDropdown?

I haven't considered any other props yet but let me check and get back to you on that. I'll hold off on releasing a new version until then.

> Also the doc changes is not reflected in the d.ts/typings

The doc in there looked Ok to me. Do you mean they are not the same verbatim?

ellinge:
Everything else is based up on the doc’s formulation so just a check so you simply just forgot.

mrchief:
I got distracted by the fact that the readme is over 10K char limit now. The all-contributors table is probably the biggest culprit. Need to rethink it now.

mrchief:
@ellinge Looking at the types, I was wondering why we are documenting props again there. Is it standard practice in the typescript community?

ellinge:
They are used by the IDE for intellisense (in case of vs code/visual studio) just as documented props in C#.

ellinge:
bild
@mrchief vs code

mrchief:
Hmmm, I suspected as much. I wonder if there's a way to DRY up this stuff without going over the top. Sometimes duplication is simpler.

PR #248, opened 2019-05-04
fix: Remove babel-runtime from distribution bundle (#248) 🔥

mrchief:
## What does it do?

Removes `babel-runtime` from the distribution bundle, reducing the size of the component.  `babel-runtime` is only needed for tests.

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,13 +1,19 @@
 {
   "presets": [["es2015", { "modules": false }], "stage-0", "react"],
-  "plugins": [
-    "transform-class-properties",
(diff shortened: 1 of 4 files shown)


ellinge:
https://bundlephobia.com/result?p=react-dropdown-tree-select@1.20.1 postreview, looks good 👍

mrchief:
We're growing but that's because we're now ***doing*** more.

I'd like to bring to single digits (gzip+min) :)

PR #251, opened 2019-05-10
refactor: Update mode to include hierarchical

mrchief:
BREAKING: hierarchical prop

`hierarchical` prop is now moved under `mode` prop.

```
// before
<DropdownTreeSelect data={data} hierarchical={true} />

// after
<DropdownTreeSelect data={data} mode="hierarchical" />
```

### DIFF
--- a/docs/src/stories/Options/index.js
+++ b/docs/src/stories/Options/index.js
@@ -45,7 +45,6 @@ class WithOptions extends PureComponent {
       showPartiallySelected,
       disabled,
       readOnly,
-      hierarchical,
     } = this.state
 
     return (
@@ -66,6 +65,7 @@ class WithOptions extends PureComponent {
               <option value="multiSelect">Multi select</option>
               <option value="simpleSelect">Simple select</option>
               <option value="radioSelect">Radio select</option>
+              <option value="hierarchical">Hierarchical</option>
             </select>
           </div>
           <Checkbox
@@ -94,7 +94,6 @@ class WithOptions extends PureComponent {
           />
           <Checkbox label="Disabled" value="disabled" checked={disabled} onChange={this.onOptionsChange} />
           <Checkbox label="Read Only" value="readOnly" checked={readOnly} onChange={this.onOptionsChange} />
-          <Checkbox label="Hierarchical" value="hierarchical" checked={hierarchical} onChange={this.onOptionsChange} />
         </div>
         <div>
           <DropdownTreeSelect
@@ -110,7 +109,6 @@ class WithOptions extends PureComponent {
             showPartiallySelected={showPartiallySelected}
             disabled={disabled}
             readOnly={readOnly}
-            hierarchical={hierarchical}
           />
         </div>
       </div>
--- a/README.md
+++ b/README.md
@@ -51,6 +51,7 @@ A lightweight and fast control to render a select component that can display hie
   - [keepOpenOnSelect](#keepopenonselect)
   - [mode](#mode)
     - [multiSelect](#multiSelect)
+    - [hierarchical](#hierarchical)
     - [simpleSelect](#simpleSelect)
     - [radioSelect](#radioSelect)
   - [showPartiallySelected](#showpartiallyselected)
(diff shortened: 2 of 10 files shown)


ellinge:
Just  checked on my mobile. Will check further later. Looks good 👍 You neee to update the story for options. 

Should we do something about showShopdown, shopDropdownAlways. 
What about oneOf(true, false, 'always') or something breaking for true/false as well. 

The search-related props (keep*) could be bundled also. Perhaps one should not go over the top though. Thoughts?

mrchief:
> The options tree/search props (keep*) could be bundled also? Perhaps one should not go over the top though with the bundling. Thoughts?

I concur. I don't see anything else as a natural candidate for bundling.

> Should be alright I we don't see anything else to bundle, but we could do that in a separate PR instead.

Agreed. The `showDropdown` bundling will be a separate PR. Thinking about `showDropdown: 'initial' | 'always'`

Good catch about updating the Options story. Will do that.

mrchief:
@ellinge Do you know if we absolutely need the semis in the `.ts` file? Prettier keeps removing them and I'm not sure if that's an error.

mrchief:
[This](https://www.typescriptlang.org/play/#src=%2F%2F%20tslint%3Adisable%3Ainterface-name%0D%0Adeclare%20module%20%22react-dropdown-tree-select%22%20%7B%0D%0A%20%20%2F%2Fimport%20*%20as%20React%20from%20%22react%22%0D%0A%0D%0A%20%20export%20type%20TreeData%20%3D%20Object%20%7C%20TreeNodeProps%5B%5D%0D%0A%0D%0A%20%20export%20interface%20DropdownTreeSelectState%20%7B%0D%0A%20%20%20%20showDropdown%3A%20boolean%0D%0A%20%20%20%20searchModeOn%3A%20boolean%0D%0A%20%20%20%20allNodesHidden%3A%20boolean%0D%0A%20%20%20%20tree%3A%20TreeNode%5B%5D%0D%0A%20%20%20%20tags%3A%20TreeNode%5B%5D%0D%0A%20%20%7D%0D%0A%0D%0A%20%20%20%20%20%20export%20interface%20TreeNode%20%7B%0D%0A%20%20%20%20%20%20%20%20label%3A%20string%3B%0D%0A%20%20%20%20%20%20%20%20value%3A%20string%2C%0D%0A%20%20%20%20%20%20%20%20checked%3F%3A%20boolean%3B%0D%0A%20%20%20%20%20%20%20%20disabled%3F%3A%20boolean%0D%0A%20%20%20%20%20%20%20%20expanded%3F%3A%20boolean%0D%0A%20%20%20%20%20%20%20%20className%3F%3A%20string%3B%0D%0A%20%20%20%20%20%20%20%20tagClassName%3F%3A%20string%3B%0D%0A%20%20%20%20%20%20%20%20isDefaultValue%3F%3A%20boolean%2C%0D%0A%20%20%20%20%20%20%20%20%5Bproperty%3A%20string%5D%3A%20any%3B%0D%0A%20%20%7D%0D%0A%0D%0A%20%20%20%20%20%20%20export%20interface%20TreeNodeProps%20extends%20TreeNode%20%7B%0D%0A%20%20%20%20%20%20%20%20%2F**%20Array%20of%20child%20objects%20*%2F%0D%0A%20%20%20%20%20%20%20%20children%3F%3A%20TreeNode%5B%5D%0D%0A%20%20%20%20%7D%0D%0A%7D%0D%0A) uses a mix of semis, commas, and no semis and doesn't seem to cause any syntax errors. According to this old [issue](https://github.com/Microsoft/TypeScript/issues/10730), semis are optional too. If so, we can let prettier loose and strip the semis. Thoughts?

docs/src/stories/Options/index.js
```diff
@@ -66,6 +65,7 @@ class WithOptions extends PureComponent {
               <option value="multiSelect">Multi select</option>
               <option value="simpleSelect">Simple select</option>
               <option value="radioSelect">Radio select</option>
+              <option value="hierarchical">Hierarchical</option>
```
mrchief: You beat me to it! And here I was wondering what the heck is he talking about - stories is already updated! 😆

ellinge:
Go ahead and strip them. I think we’ll be just fine and concistent.

ellinge:
Still works great within vs code if that was what you intended me to verify after dismissing my review.

mrchief:
> Still works great within vs code if that was what you intended me to verify after dismissing my review.

There's a GitHub setting that auto dismisses previous reviews if you push new commits. :)

PR #253, opened 2019-05-18
refactor: Consolidate showDropdown props

mrchief:
BREAKING CHANGE: **`showDropdown`** & **`showDropdownAlways`** props are merged

`showDropdown` and `showDropdownAlways` props are now moved under `showDropdown` prop.

```
// before
<DropdownTreeSelect data={data} showDropdown />

// after
<DropdownTreeSelect data={data} showDropdown="initial" />
```

**showDropdown: always**

```
// before
<DropdownTreeSelect data={data} showDropdownAlways />

// after
<DropdownTreeSelect data={data} showDropdown="always" />
```


src/a11y/a11y.test.js
```diff
@@ -48,15 +48,15 @@ test('regular tree has no a11y exceptions', async t => {
   const { tree } = t.context
   const component = mountToDoc(
     <div>
-      <DropdownTreeSelect data={tree} showDropDown texts={{ label: 'test' }} />
```
mrchief: @ellinge Regarding the failing tests: `showDropDown` (note the casing) was never a valid prop. So these tests were passing earlier (I think that's the reason why). Including the correct `showDropdown` prop throws a weird a11y error which I've no idea why.

mrchief: Guess I figured it out. https://www.w3.org/TR/wai-aria-1.1/#list & https://www.w3.org/TR/wai-aria-1.1/#tree do not support `aria-readonly`. 

Also, in case of `simpleSelect`, the `role=list` seems more fitting as [`listbox`](https://www.w3.org/TR/wai-aria-1.1/#listbox) implies multiple items can be selected:

> A widget that allows the user to select one or more items from a list of choices.

mrchief: Scratch that. According to https://www.w3.org/TR/wai-aria-1.1/#select, listbox is the correct attribute. Damn, this Aria stuff is confusing.

The only failing test is for radio select now. Gonna look into that next.

ellinge: Added a role="radiogroup" on the dropdown content for radioselect now

ellinge: Based on https://www.w3.org/TR/2016/WD-wai-aria-practices-1.1-20160317/examples/radio/radio.html

src/tree/index.js
```diff
@@ -128,12 +128,18 @@ class Tree extends Component {
...
+    const { mode } = this.props
+
+    const attributes = {
+      /* https://www.w3.org/TR/wai-aria-1.1/#select
+       * https://www.w3.org/TR/wai-aria-1.1/#tree */
+      role: mode === 'simpleSelect' ? 'listbox' : 'tree',
+    }
+
+    if (/multiSelect|hierarchical/.test(mode)) {
+      attributes['aria-multiselectable'] = true
```
ellinge: The reason for 'true' : 'false' before is because that forces the values to be outputted, not only on true (boolean true).

Not sure if we should keep that (does not seem to cause any a11y-errors when running tests)
```
    return {
      /* https://www.w3.org/TR/wai-aria-1.1/#select
       * https://www.w3.org/TR/wai-aria-1.1/#tree */
      role: mode === 'simpleSelect' ? 'listbox' : 'tree',
      'aria-multiselectable': /multiSelect|hierarchical/.test(mode) ? 'true' : 'false',
    }
```

src/a11y/a11y.test.js
mrchief: > What a miss on my part with the casing. Wouldn't have happened with typescript 😁

I see what you did there! :)

How would TS know it's not a valid prop?

src/tree/index.js
mrchief: I guess this needs revisiting. It was complaining about them earlier but that could have been due to something else.

src/a11y/a11y.test.js
ellinge: It would not compile and the IDE would tell me straight away.
Well at least as long as the props interface wasn’t simply `any` or allowing custom props through a any indexer.

Probably can install something in VS Code to check against propTypes as well.
https://marketplace.visualstudio.com/items?itemName=OfHumanBondage.react-proptypes-intellisense

mrchief: > Well at least as long as the props interface wasn’t simply any or allowing custom props through a any indexer.

I think you can enforce something similar via prop types too but not sure. Way back, we built https://github.com/dowjones/react-json-schema-proptypes which would enforce that too (it used a JSON schema and provided prop types in return). 

React components and JS by nature, do not explicitly disallow any custom props. In fact, disallowing them is more work and allowing them sometimes has unintended benefits to the user. In cases like this though, it bites us back. 

But to be fair, it escaped my eyes too. I did see those errors before but I simply thought it was due to bad merge resolutions. So doubly bad for me ignoring things twice. :)

That extension seems like a good one. Gonna try it.

src/tree/index.js
ellinge: We need them to be strings for them to be emitted on false also.

mrchief: I think the reason a11y tests pass is because of this:

from https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/listbox_role

> **aria-multiselectable**
Include and set to true if the user can select more than one option. If set to true, every option should have a the aria-selected attribute included. If false or omitted, only the currently selected option, if any option is selected, needs the aria-selected attribute, and it must be set to true.

So going by this, it's ok to omit for single select cases.

PR #255, opened 2019-05-20
feat: Ability to dedup tags with same values

### LINKED ISSUE #185: Selecting all nodes with same name renders duplicate tags in textbox
When there are multiple nodes with the same name, on selection of two or more of these nodes, duplicate tags appear in the box. Is there any way we can customize the component so that only unique tags appear?

### LINKED ISSUE #254: Allow unique tags in Input for data with duplicate nodes
In my use-case I have duplicate nodes in data tree, i.e., nodes with same label & value at different nodes in tree. I use it in hierarchical mode and onChange select all nodes with same value when anyone is selected. I have requirement to show single tag in input.
PS: This is similar to https://github.com/dowjones/react-dropdown-tree-select/issues/185

A new prop to DropdownTreeSelect called dedupTags that would filter duplicate tags by value.

**Additional context**
Ex: We have 'Marmara University' duplicated in the data
Existing/Actual behavior is to show Marmara University twice. As shown below
image

Expected behavior is to show single tag for Marmara University selection
image

dpkm95:
## What does it do?
Adds ability to dedup tags with same value.

## Fixes # (issue)
Closes #254 
Fixes #185

## Type of change

# Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] Added myself to contributors table
- [ ] I have updated `docs/bundle.js` file (ran `yarn build:docs` and checked in the new `bundle.js` file)


dpkm95:
I have raised the pr for discussion, will update tests and documentation soon. Please give feedback on anything else that I've missed.

mrchief:
@dpkm95 Thanks for sending the PR.

As I see the changes and look at your description, I've mixed feelings about this. There are 2 issues that stand out to me:

 - Having 2 identical tags next to each other looks like a data entry error. What is the use case that would permit 2 labels to have the same exact text and still be valid as 2 different entries?

 - Secondly, filtering it out at the `input` level is a bit deceiving, IMHO. You're treating the symptoms, not the root cause.

#185 was not a bug, rather it **_seemed like one_** due to the underlying data which had identical nodes (albeit with different parents). That case is better solved by 

  a) removing dup nodes within data 
  b) adding some sort of tag/marker in the data to distinguish its pedigree 
  c) use css to distinguish 
  d) any other valid UX/design based on your business/needs
  e) introduce some sort of hook/event that lets you format the label - again it isn't strictly WYSIWYG and introduces additional complexity so we need very good reasons for doing so.

dpkm95:
@mrchief Thanks for the analysis. I agree with you using `tagClassName: dupNode` for duplicates seems to be right way.
PS:
`.dupNode {
    display: none;
}`

PR #259, opened 2019-05-22
fix: Avoid jarring on paging

### LINKED ISSUE #257: Option to change pageSize
I have a use case with a list of ~250 elements and the pageSize of 100 causes jarring when selecting an item from the end of the list.

Ability to set custom `pageSize` value.

ellinge:
Avoids changing scroll if the current active descendant is the same for the tree (on prop updates on tree).
Also fixes jarring issue when selecting nodes on paging (index > 100/pagesize).

Partially fixes #257

## Type of change
- [x] Bug fix


src/index.keyboardNav.test.js
```diff
@@ -161,3 +161,47 @@ test('should set current focus as selected on tab out for simpleSelect', t => {
   triggerOnKeyboardKeyDown(wrapper, ['ArrowDown', 'ArrowRight', 'ArrowRight', 'Tab'])
   t.deepEqual(wrapper.state().tags[0].label, 'ccc 1')
 })
+
+test('should scroll on keyboard navigation', t => {
```
mrchief: Love these tests! 👍

mrchief:
Thanks for jumping in on this!

src/index.keyboardNav.test.js
ellinge: I need to add one ensuring scroll is intact on regular use also.
Everthing gets a bit harder without a real dom though 😆

ellinge: Commenting the following line https://github.com/dowjones/react-dropdown-tree-select/blob/f7b7dd5339f4879207587c5a0d9ea3884154786d/src/tree/index.js#L53 fails the newly added test.

src/tree/index.js
```diff
@@ -134,7 +135,7 @@ class Tree extends Component {
       /* https://www.w3.org/TR/wai-aria-1.1/#select
        * https://www.w3.org/TR/wai-aria-1.1/#tree */
       role: mode === 'simpleSelect' ? 'listbox' : 'tree',
-      'aria-multiselectable': /multiSelect|hierarchical/.test(mode),
+      'aria-multiselectable': /multiSelect|hierarchical/.test(mode).toString(),
```
mrchief: Do we need the `toString()` or is it from before our convo on the other PR?

ellinge: Yes. It will only be appended on true otherwise.

mrchief: Right, but omitting that is ok. See https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/listbox_role

mrchief:
This isn't released yet. I gotta straighten out quirks with the release process. But when it does, it won't be a major version update.

mrchief:
Released now - v2.0.1

PR #265, opened 2019-06-14
fix: Remove resetting of dropdown on init props (#265) 🐛

ellinge:
## What does it do?
Keeps dropdown open if updating with new props. Only updates/sets state open  if new props causes the dropdown to show (a.k.a initial/always)

#253 introduced resetting of state for shopDropdown on new props:
https://github.com/dowjones/react-dropdown-tree-select/blob/297ed42672a67b60858ab48192cd6717892fc71f/src/index.js#L82

## Type of change
- [x] Bug fix


mrchief:
Should we add a test for this?

ellinge:
> Should we add a test for this?

Tests added, Btw, offtopic, but noticed this: 
https://unpkg.com/react-tree-dropdown-@2.1.3-beta.0/dist/react-dropdown-tree-select.js
```
            nodeRenderer: i.a.func,
            iconRenderer: i.a.func,
            tagRenderer: i.a.func,
```

mrchief:
> Tests added, Btw, offtopic, but noticed this:

Interesting! :) 

Also, there are quite a few:

image

They could have used patch-package instead of going thru all the trouble :)

mrchief:
Oh, I remember now: https://github.com/dowjones/react-dropdown-tree-select/pull/162

ellinge:
I used PRs and temporary npm git installs instead, but perhaps this would have been faster 🤔😆

mrchief:
Yeah, git installs will work too! I've found that patch-package can help in cases where your npm is behind an enterprise proxy or a private repo. Or if your build system doesn't understand/support that syntax. Plus yarn syntax for installing from PRs is a bit different than npm so that's another inconvenience with git installs. 🤷‍♂ 

In the end, use what works best for you. I'm just happy to have so many choices!

PR #270, opened 2019-06-27
feat: Add support for custom search filtering on nodes (#270) ✨

### LINKED ISSUE #269: Add possibility to use custom predicate/filter when searching nodes
We have found a use case where we want to have "shortcut"-match on items which is not displayed as a label on the item. How about an optional additional field, a.k.a, hidden label, on the passed node data which the searchterm also is filtered against?

ellinge:
## What does it do?

Adds an additional property, searchPredicate, which is used when filtering/searching for nodes.
Can be used to have acronyms and other metadata, which should cause a match/hit, or to implement a fuzzy search.

Fixes #269

## Type of change

- [x] New feature

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,7 @@ A lightweight and fast control to render a select component that can display hie
     - [always](#always)
   - [form states (disabled|readOnly)](#formstates)
   - [id](#id)
+  - [searchPredicate](#searchPredicate)
 - [Styling and Customization](#styling-and-customization)
   - [Using default styles](#default-styles)
   - [Customizing with Bootstrap, Material Design styles](#customizing-styles)
@@ -267,7 +268,6 @@ Data for rendering the tree select items. The object requires the following stru
   actions,        // optional: An array of extra action on the node (such as displaying an info icon or any custom icons/elements)
   dataset,        // optional: Allows data-* attributes to be set on the node and tag elements
   isDefaultValue, // optional: Indicate if a node is a default value. When true, the dropdown will automatically select the node(s) when there is no other selected node. Can be used on more than one node.
-  hiddenLabel,    // optional: Checkbox hidden label, used for search/filtering
   ...             // optional: Any extra properties that you'd like to receive during `onChange` event
 }
 ```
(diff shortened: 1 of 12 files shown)


src/tree-manager/index.js
```diff
@@ -50,7 +50,7 @@ class TreeManager {
     const matches = []
 
     const addOnMatch = node => {
-      if (node.label.toLowerCase().indexOf(searchTerm) >= 0) {
+      if ((node.label + ' ' + node.hiddenLabel).toLowerCase().indexOf(searchTerm) >= 0) {
```
mrchief: Shouldn't this be better off by matching on individual props rather than concatenating them, like:

```
if (node.label.toLowerCase().indexOf(searchTerm) >= 0 || node.hiddenLabel.toLowerCase().indexOf(searchTerm) >= 0)
```

It's a little more verbose but seems more true to its purpose and avoids creating extra strings

ellinge: It's a function now

src/tree-manager/index.js
```diff
@@ -278,6 +275,19 @@ class TreeManager {
...
   }
+
+  _getAddOnMatch(matches, searchTerm) {
+    let isMatch = (node, term) => node.label.toLowerCase().indexOf(term) >= 0
+    if (this.searchPredicate && typeof this.searchPredicate === 'function') {
+      isMatch = this.searchPredicate
+    }
+
+    return node => {
+      if (isMatch(node, searchTerm) === true) {
```
mrchief: Redundant true comparison?

src/tree-manager/index.js
```diff
@@ -278,6 +275,19 @@ class TreeManager {
 
     return keyboardNavigation.handleToggleNavigationkey(action, prevFocus, readOnly, onToggleChecked, onToggleExpanded)
   }
+
+  _getAddOnMatch(matches, searchTerm) {
+    let isMatch = (node, term) => node.label.toLowerCase().indexOf(term) >= 0
+    if (this.searchPredicate && typeof this.searchPredicate === 'function') {
```
mrchief: Just a `typeof` should suffice, no?

src/tree-manager/index.js
ellinge: In the custom function they could return anything they wanted that could be evaluated to true? Should we let them? I mean return "hello world" would we valid?

src/tree-manager/index.js
ellinge: Yeah it should 👍

src/tree-manager/index.js
mrchief: We kinda don't care... Same concept as `array.filter` or any other native matcher. We just care that the value is `truthy` since the return value is not being used anywhere.

README.md
```diff
@@ -60,6 +60,7 @@ A lightweight and fast control to render a select component that can display hie
     - [always](#always)
   - [form states (disabled|readOnly)](#formstates)
   - [id](#id)
+  - [searchPredicate](#searchPredicate)
```
mrchief: This needs to be all lowercase like `(#searchpredicate)`. GH doesn't link otherwise. Gets me everytime. :)

But don't submit another PR. I'll be fixing this in a doc update. Experimenting with a new readme layout that doesn't need so much scrolling.

Something like

image

PR #272, opened 2019-06-30
docs: Compact Readme

mrchief:
## What does it do?

An attempt to reduce the amount of scrolling in the README

## Type of change

- [ ] New feature (non-breaking change which adds functionality)
- [X] This change requires a documentation update

# Checklist:

- [X] Added tests that prove my fix is effective or that my feature works

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,41 +26,20 @@
 
 A lightweight and fast control to render a select component that can display hierarchical tree data. In addition, the control shows the selection in pills and allows user to search the options for quick filtering and selection. Also supports displaying partially selected nodes.
 
+## At a Glance
+
+
+| [Demo](#example) | [Props](#props) | Events | [Keyboard navigation](#keyboard-navigation) |
+|:----------------:|:-------------------------------------------------------:|:------------------------------------:|:-------------------------------------------:|
+| [**Modes**](#mode) | [**Styling and Customization**](#styling-and-customization) | [**Doing more with HOCs**](/docs/HOC.md) | [**Custom Filtering**](#searchPredicate) |
+
 ## Table of Contents
 
 - [Screenshot](#screenshot)
 - [Demo](#example)
-  - [Vanilla (no framework)](#vanilla-no-framework)
-  - [With Bootstrap](#with-bootstrap)
-  - [With Material Design](#with-material-design)
-  - [As Single Select](#as-single-select)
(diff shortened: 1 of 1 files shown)


ellinge:
The icons have some ”hamburger menu icon” between them. Intended? Mobile iOS.
On mobile android there’s like a fat line between them. How about keeping them on the same line and a simple slash between the icons?

mrchief:
> The icons have some ”hamburger menu icon” between them. Intended? 

Ha, no! Putting a simple pipe breaks markdown so I tried some similar looking Unicode character. Probably that code point represents something else on Mobile OS's font.

I think I'll get rid of them and stick to simple text labels.

> Where is events intended to be linked>

Was thinking about be moving event props together and giving them a subhead. Then we can link them up.

I'm pushing some changes now, let know what you think.

ellinge:
F2CB7A81-A86E-401E-AA47-56C91B4B5D80
Should be same styling on the sublinks. Try escaping the pipes instead of the char you are using now?

ellinge:
Added a commit now, easier than suggestions. Since github does not support a headerless table I added ** to the "first row" also. I escaped the pipes as well. I think a simple / should be good though instead.

mrchief:
>  Try escaping the pipes instead of the char you are using now?

Duh! Thought I fixed that in my recent changes. 

> Added a commit now, easier than suggestions.

Agreed. Thanks for the commit.

> I think you would have more control using html-table instead of markdown but you would have to convert the cell content also then I think

Yeah, dabbled with that idea as well. It's more verbose so adds to overall bytes. The readme stands more than 10K bytes right now so some git diff editors think its a binary file. Apart from that, I think 10K bytes is probably a smell. The avg read time for current readme is ~22 minutes. I'm trying to see there is something we can do about reducing that time. A short readme is certainly more palatable but I also don't want to lose any vital information.

I'm ok with losing a little bit of vanity for less typing. As a workaround, I played with the idea of bolding subsequent rows. It's not ideal but then raw HTML tables are even uglier to work with. 🤷‍♂ 

Feel free to submit any other ideas you may have. E.g., I'm still not sure if ditching the ToC entirely is a good move or not.

ellinge:
What about splitting the readme into separate files? A general quickguide and separate readmes per section for deep diving?

PR #275, opened 2019-07-05
fix: Adjust aria labels to reference selections and read Remove

ellinge:
## What does it do?

We had some, hard to interpret, feedback about aria/labels from a third party checking accessibility.
So, I installed [NVDA](https://www.nvaccess.org/) to check more exactly how a screen reader interprets all the aria-attributes the component outputs. Anyway, I noticed that the aria-label="Remove" was not spoken. I tried with Firefox and Chrome (and [ChromeVox ](https://chrome.google.com/webstore/detail/chromevox/kgejglhpjiefppelpmljglcjbhoiplfn?hl=sv) with Chrome).

Also, the selections are not read when focusing on the whole dropdown/trigger, so added them as additional labelled by references. It then reads the label + the tags selected when focusing on the dropdown. It’s how the [example on W3](https://www.w3.org/TR/wai-aria-practices/examples/listbox/listbox-collapsible.html ) works as well. It reads ~ “Choose an element Neptunium menu button sub menu”.

## Type of change

- [x] Bug fix


src/a11y/index.js
```diff
@@ -1,9 +1,19 @@
-export function getAriaLabel(label) {
-  if (!label) return undefined
+export function getAriaLabel(label, additionalLabelledBy) {
+  const attributes = getAriaAttributeForLabel(label)
 
-  if (label && label[0] === '#') {
-    /* See readme for label. When label starts with # it references ids of dom nodes instead.
-      When used on aria-labelledby, they should be referenced without a starting hash/# */
+  if (additionalLabelledBy) {
+    attributes['aria-labelledby'] = ((attributes['aria-labelledby'] || '') + ' ' + additionalLabelledBy).trim()
```
mrchief:
```suggestion
    attributes['aria-labelledby'] = `${attributes['aria-labelledby'] || ''} ${additionalLabelledBy}`.trim()
```

src/a11y/index.js
```diff
@@ -1,9 +1,19 @@
...
+
+function getAriaAttributeForLabel(label) {
+  if (!label) return {}
+
+  /* See readme for label. When label starts with # it references ids of dom nodes instead.
+    When used on aria-labelledby, they should be referenced without a starting hash/# */
+  if (label[0] === '#') {
     return { 'aria-labelledby': label.replace(/#/g, '') }
   }
   return { 'aria-label': label }
```
mrchief:
```suggestion
  return { 'aria-label': label.replace(/^#/g, '') }
```

We can probably forgo the if check too and just do the replace all the time. The comment is descriptive enough.

src/index.js
```diff
@@ -298,12 +298,12 @@ class DropdownTreeSelect extends Component {
             { 'radio-select': mode === 'radioSelect' }
           )}
         >
-          <Trigger onTrigger={this.onTrigger} showDropdown={showDropdown} {...commonProps}>
+          <Trigger onTrigger={this.onTrigger} showDropdown={showDropdown} {...commonProps} tags={tags}>
```
mrchief: ` tags={tags}` Are using tags for populating aria attributes? If so, then:

 - why does it need to know since the trigger is not rendering the tags
 - how about pulling up the `tags > label logic` & pass the labels?

src/tag/index.js
```diff
@@ -41,23 +41,22 @@ class Tag extends PureComponent {
...
     const isDisabled = readOnly || disabled
-    const onClick = !isDisabled ? this.handleClick : undefined
-    const onKeyDown = !isDisabled ? this.onKeyDown : undefined
-    const onKeyUp = !isDisabled ? this.onKeyUp : undefined
 
     return (
-      <span className={cx('tag')} id={tagId}>
-        {label}
+      <span className={cx('tag')}>
+        <span id={tagId}>{label}</span>
```
mrchief: This could be potentially a breaking change as it alters the HTML markup rendered by previous versions.

ellinge: I thought it to be very unlikely that someone would have any selectors affected by this change. I weighted it against setting aria-label={label} which also solves the problem with ”remove” being spoken when focusing the trigger node.

src/index.js
ellinge: Well the tags are rendered/labeled/id:ed in the child node/Input. So it would be if I would ref them or bundle them (Input and Trigger) more "together". Instead of passing the input as a child one could render the Input within the trigger-component instead? Or what are you suggesting?

src/a11y/index.js
ellinge: The id check causes different aria-attributed to be outputted. Aria-label is used for text and aria-labelledby is used for references/ids.

src/tag/index.js
mrchief: Yeah, I'm not against changing the HTML but then it'd be a major version bump (to be safe, just in case anyone is relying on it). I think, setting `aria-label={label}` is probably preferable.

src/a11y/index.js
mrchief: Ah, sorry, should have looked closely. What about the regex `label.replace(/^#/, '')`? Do you still need to do the global replace?

src/index.js
mrchief: The trigger is a separate component, it doesn't need to know anything about input. The reason there is a parent-child relationship right now is for backward compatibility. I was using a 3rd party component which outputted the current HTML structure. When I replaced that 3rd party with in-house, I kept the same structure. 

The main component (index.js) is the owner and currently houses most business logic. Individual components only concern themselves with presentational logic. That was roughly the idea/organization that I started with. 

We could just break free of legacy baggage and redesign this in a manner where Trigger doesn't appear to be Input's parent. 

If we do not want to cause a major change, then we could calculate aria attributes for tags and pass them to tags as well as Trigger component. Thing is if we go this route, we're breaking the encapsulation of presentational logic of a component within the component (aria attributes of tags being calculated outside). That'd be acceptable if we do it for all components - so aria attributes are calculated outside of each component and then passed on to them as props. 

I'm not sure which one is better, but now you know a bit of context. Like always, all suggestions welcome! :)

src/a11y/index.js
ellinge: In theory one could have more than one label describing the content, e.g
#refinelabel #categorylabel, so yes it is needed.

mrchief: Not sure I follow. DOM ids start with a hash so you’re trying to replace them: `#someId > someId`. Correct? But the regex you have will replace all hashes like `#some#Id > someId`. Is this intended?

ellinge: You can have multiple label references/ids. If you want to be picky one could replace the regex with label.substring(1).replace(/[ ]#/g but it would still need to be global. If you start of with a hash you probably are not writing a text label though.

mrchief: > But hash is actually allowed in a id in html5 so I guess one should allow it even if highly unlikeley.

I'm not sure how we can support it reliably without introducing more logic to correctly distinguish between id hashes and hashes that are part of the label. 

But when you say 

> You can have multiple label references/ids.

do you mean something like `#myDivOne #myDivTwo`? In that case, `label.substring(1).replace(/ #/g, '')` will be able to handle it without replacing hash in the middle of a label.

src/index.js
ellinge: Not sure how I could do this simpler that it is with its current structure and did not have the intention of rewriting the component, just fix so that the correct references are set between the "focus node" and its selected tags. And those are generated two steps down (the ${_id}_tag of the tags based on passed state.tags). 

The only idea would be to have a function which "bubbles" up the generated tag ids or references to pass to the parent. But not sure how much of a difference that would make...

I made a shared function for tag id generation so that's done in a single way at least.

mrchief: Hear ya… Let's go with this now and revisit this later.

__snapshots__/src/index.test.js.md
```diff
@@ -177,12 +180,15 @@ Generated by [AVA](https://ava.li).
         className="dropdown"
       >
         <Trigger
+          clientId="rdts"
           disabled={true}
           onTrigger={Function {}}
           showDropdown={false}
+          tags={[]}
```
mrchief: Shouldn’t there be tags here instead of an empty array?

src/index.test.js
```diff
@@ -290,3 +290,19 @@ test('adds aria-label when having label on search input', t => {
...
+  t.deepEqual(wrapper.find('.dropdown-trigger').prop('aria-labelledby'), 'hello world rdts-0_tag')
+  t.deepEqual(wrapper.find('.dropdown-trigger').prop('aria-label'), undefined)
+})
+
+test('appends selected tags to aria-labelledby with text label', t => {
+  const { tree } = t.context
+  tree[0].checked = true
+  const wrapper = mount(<DropdownTreeSelect id="rdts" data={tree} texts={{ label: 'hello world' }} />)
+  t.deepEqual(wrapper.find('.dropdown-trigger').prop('aria-labelledby'), 'rdts_trigger rdts-0_tag')
+  t.deepEqual(wrapper.find('.dropdown-trigger').prop('aria-label'), 'hello world')
```
mrchief: We don’t wanna rely on snapshots for these checks (this and above one)?

src/tree-manager/keyboardNavigation.js
```diff
@@ -151,7 +151,7 @@ const getNextFocusAfterTagDelete = (deletedId, prevTags, tags, fallback) => {
 
   index = tags.length > index ? index : tags.length - 1
   const newFocusId = tags[index]._id
-  const focusNode = document.getElementById(getTagId(newFocusId))
+  const focusNode = document.getElementById(`${getTagId(newFocusId)}`)
```
mrchief: What’s the difference (I mean this vs `document.getElementById(getTagId(newFocusId))`)?

mrchief: Ah, I see you reverted. Nevermind.

__snapshots__/src/index.test.js.md
ellinge: Not sure I follow. It’s the same state.tags as sent to the input, which is empty?

src/index.test.js
ellinge: Do you want me to switch to snapshots instead?

__snapshots__/src/index.test.js.md
mrchief: Seems so.

src/index.test.js
mrchief: Explicit testing is OK for a handful of assets. You can also express the intent by using inline snapshots but I’m not sure if Ava supports them yet. Let’s leave these as is for now.

__snapshots__/src/index.test.js.md
ellinge: https://github.com/dowjones/react-dropdown-tree-select/blob/8eefed776959458a657c90b48d0360252fb1eb63/src/index.test.js#L28
Do you want me to set a node as checked/selected?

mrchief: I was hoping to look at this last weekend but didn’t get a chance. I think I’m not understanding the intent here. 

What I assumed previously is that this change is supposed to populate tags as aria attributes in the rendered markup. The snapshot shows empty so I thought that was a mistake. From your last comment, it seems that happens only when you check a node. If so, then I think we there should be a test to verify that behavior (which I think you already have added). If all of that is true, then there is nothing to do here and we can mark this as resolved.

ellinge: The snapshot was changed because the input needs info about which nodes are checked/tagged. No intent behind this specific tests. The output (aria-attributed) is verified in other tests.

PR #279, opened 2019-07-24
feat: Add custom keys for label and value

### LINKED ISSUE #276: Add custom name for label and value
Hello!
Is it possible to add the ability to set a custom name for the label and the value from the props?

SergeyWebDeveloper:
## What does it do?

Add opportunity establish custom field property from the input data array from which it will be taken `label` and `value`

## Fixes # (issue)

"Fixes #276"

## Type of change

# Checklist:


mrchief:
@SergeyWebDeveloper Thanks for sending this great PR! I now understand what you're trying to do. May I suggest an HoC approach to solve this? https://codesandbox.io/s/custom-labelvalue-with-react-dropdown-tree-select-ybevy

SergeyWebDeveloper:
@mrchief Your solution is excellent, but I wanted this opportunity to be available in the library.
Using `deep-rename-keys` is unnecessary recursion and unnecessary traversal of the tree, it is much more productive to do all this in the library. Correct me if my thoughts are wrong. Thanks

SergeyWebDeveloper:
How long should I wait for an answer?

mrchief:
Sorry, been busy. I'll let you know tonight.

mrchief:
@SergeyWebDeveloper thanks for sending this PR. I checked your changes. You're right about the extra recursion. Whether that's a performance impact or not depends on one's use case. 

However, I'm not sure if I want to support _mapping of one data node to another_ as a core feature. One of the design philosophies behind this component is to keep things as simple as possible. Adding the transformation logic creates more moving parts than needed, for the component, to do its job. It makes any future contributors' job harder by increasing the code complexity. Think about this - extending the same logic and reasoning, any core prop can be customized. So where does one draw the line? And how does one ensure that the component does its job properly? Imagine the complexity this would add to testing, in order to ensure that not only the transformation bits do their job properly, but the rest of the component functions as expected with every possible permutation of such transformation.

Frontend components should not be in charge of massaging data, their concern should be their presentation. In the docs, you'll see that there are many HOC based examples. Each one of them is as good a candidate for inclusion in the core component. But are they absolutely necessary? No. Are they going to be helpful to _everyone_? Perhaps not; which is why they live on the outside.

PR #283, opened 2019-08-26
feat: Apply search after update props

malegender:
## What does it do?

When the `data` property changes, the search is not applied.
Needed when children load dynamically(when expanding node)

## Type of change

# Checklist:


mrchief:
> When the data property changes, the search is not applied.

Agreed.

> Needed when children load dynamically(when expanding node)

Could you expand on this? I can't see this as an existing bug. E.g, here's a screen capture from the Large Tree demo:

4r3NoYpVMw

Issue #303, opened 2019-11-05
Warning message from 16.10.0 version of 'react'.


darkfoxs96:
Warning message from 16.10.0 version of 'react'.

```
Warning: react-dom.development.js:12427 componentWillMount has been renamed, and is not recommended for use. See https://fb.me/react-unsafe-component-lifecycles for details.
```
```
Warning: react-dom.development.js:12427 componentWillReceiveProps has been renamed, and is not recommended for use. See https://fb.me/react-unsafe-component-lifecycles for details.
```

![image](https://user-images.githubusercontent.com/35303397/68198390-fb894780-ffcc-11e9-873c-8890520d19f8.png)

YarBez:
+1 same problem

mrchief:
It's [not a problem](https://github.com/dowjones/react-dropdown-tree-select/issues/292#issuecomment-535050878) :)

PR #304, opened 2019-11-05
For #303. Fix warning message from 16.10.0 version of 'react'.

darkfoxs96:
https://github.com/dowjones/react-dropdown-tree-select/issues/303

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -96,7 +96,7 @@ class DropdownTreeSelect extends Component {
     }
   }
 
-  componentWillMount() {
+  UNSAFE_componentWillMount() {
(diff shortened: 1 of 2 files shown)


mrchief:
I'm aware of it and I don't like it either. There are 2 issues here though:

 - this was introduced in React 16.9 and adding this change means we need to update the peer depdendency constraint as it is no longer just 16+ (has to be 16.9+). It should be OK but feels like a forced upgrade (to React 16.9) for the consumers of this component.
 - You'd still see the warning as one the dependencies of this project needs to upgrade as well (react-infinite-scroll-component)

PR #309, opened 2019-11-23
feat: render the search input inside the dropdown

### LINKED ISSUE #308: render the search input inside the dropdown. not on the toggle button
I need to render the search box inside the dropdown, not on the main button.

a configuration option allows us to choose where to render the search button

**Describe alternatives you've considered**
Fork the repo!.

**Additional context**
This is the design i am trying to achieve.
image

osmancode:
## What does it do?

Allows us to show/hide the search input and change the location from the dropdown toggle to dropdown content.
## Type of change

# Checklist:


osmancode:
@mrchief not really sure how to bypass the codeclimate issue. it is not a code duplicate!.

mrchief:
Thanks for sending the PR. Let me look it over and get back to you.

> @mrchief not really sure how to bypass the codeclimate issue. it is not a code duplicate!.

CC is not perfect and it does flag some identical blocks as false duplicates. If it's a genuine false alert, I can override it so this shouldn't be an issue.

src/input/index.css
```diff
@@ -1,25 +1,10 @@
-.node > label {
-  cursor: pointer;
-  margin-left: 2px;
+.search {
```
mrchief: This would constitute a breaking change as it alters the HTML and CSS structure. Not this particular one but the PR as a whole. There are a few ways to mitigate it:

 - This feature should be opt-in, which it is by the looks of it. So if the user doesn't opt-in, nothing should change for them. That means it should render the same HTML and CSS structure.
 - The other way is to make this a breaking change, which means the structure is altered permanently and we provide a migration path for old users. 

I'd prefer the former but not if it complicates code structure or adds bloat (to testing, performance, maintenance etc.).

Let me see if I can fiddle with it but if you have considered these and have more insights to share, please let me know.

src/input/index.js
```diff
@@ -38,6 +20,7 @@ class Input extends PureComponent {
     disabled: PropTypes.bool,
     readOnly: PropTypes.bool,
     activeDescendant: PropTypes.string,
+    searchInputLocation: PropTypes.oneOf(['dropdownToggle', 'dropdownContent']),
```
mrchief: How about we make this a bool? Something like `inlineSearchInput: true|false`, which is false by default. That way, users don't have to do anything if they desire existing behavior. And then structure the rest of the code based on that boolean?

src/input/index.test.js
```diff
@@ -6,13 +6,7 @@ import toJson from 'enzyme-to-json'
...
 import Input from './index'
 
-test('renders tags', t => {
-  const tags = [{ _id: 'i1', label: 'l1' }, { _id: 'i2', label: 'l2' }]
-  const wrapper = toJson(shallow(<Input tags={tags} />))
-  t.snapshot(wrapper)
-})
-
-test('renders input when no tags are passed', t => {
+test('renders input', t => {
```
mrchief: We'd need the existing tests to pass as is; again goes back to the opt-in based designs. So removing them completely defeats their purpose. We'd need new tests for the added features too.

types/react-dropdown-tree-select.d.ts
```diff
@@ -97,6 +97,10 @@ declare module 'react-dropdown-tree-select' {
     id?: string
     /** Optional search predicate to override the default case insensitive contains match on node labels. */
     searchPredicate?: (currentNode: TreeNode, searchTerm: string) => boolean
+    /** Whether to show or hide the SearchInput */
+    showSearchInput?: boolean
```
mrchief: Why would this be a bool? It can be hidden via CSS easily.

src/input/index.test.js
osmancode: Since i renamed the Input to Tags, and the Input now only renders the Input. so the tests have been changed but the outputted html, css in still the same. 
I cant reuse the Input in two different locations without putting it in a separate component.

osmancode:
@mrchief  what do u think with the updated PR?

mrchief:
@osmancode The updated PR looks great at a glance. I'll try to dig deep into this soon.

src/tags/index.js
```diff
@@ -0,0 +1,53 @@
...
+    children: PropTypes.node,
+  }
+
+  render() {
+    const { tags, onTagRemove, texts = {}, disabled, readOnly, children } = this.props
+    return (
+      <ul className={cx('tag-list')}>
+        {getTags(tags, onTagRemove, readOnly, disabled, texts.labelRemove)}
+        {children && <li className={cx('tag-item')}>{children}</li>}
+        {!children && (!tags || (tags && tags.length === 0)) && (
```
mrchief: > !tags || (tags && tags.length === 0)

Seems like a bit of excessive branching. Can we use ES6 defaults and eliminate branches or better yet, eliminate it altogether since it seems redundant. 

```jsx
{!children && (!tags || (tags && tags.length === 0)) && (
  <li className={cx('tag-item')}>
    <span className={cx('placeholder')}>{texts.placeholder || 'Choose...'}</span>
  </li>
)}
```

If no children, you'd render the placeholder regardless of tags or no tags, no?

I was wondering if we could do something like:

```jsx
const lastItem = children || <span className={cx('placeholder')}>{texts.placeholder || 'Choose...'}</span>
return (
  <ul className={cx('tag-list')}>
    {getTags(tags, onTagRemove, readOnly, disabled, texts.labelRemove)}
    <li className={cx('tag-item')}>{lastItem}</li>
  </ul>
)
```

and maybe even inline `getTags` since there's no good reason for it to exist outside (it was done earlier to reduce code complexity when it was inside `Input`'s `render` block).

Thoughts?

osmancode:
@mrchief i have granted you push permissions over the my repo.

mrchief:
@osmancode  were you able to work on these changes?

osmancode:
@mrchief i would like to but unfortunately i will be very busy for the next two months. do u think you will be able to finalize the PR?

mrchief:
I'll try to... just didn't wanna duplicate efforts if you're already on it.

mrchief:
@osmancode I think stale bot closed this and if that's the case, I'd need to check its settings as it wasn't supposed to close `wip` PRs.

PR #318, opened 2019-12-19
feat: Adds selectable option

### LINKED ISSUE #306: Label-only nodes
I found this use case where a parent node shouldn't be selectable, but actually be just a label. Sort of serving a purpose of only grouping descendants and inform what that group is.

Screen Shot 2019-11-19 at 11 32 11
I think this would be a `mode`, or a new prop where you can set it as `label_only`. 
The checkbox (or radio) itself would disappear, the node would still be expandable, but it'd only keep the label.

**Describe alternatives you've considered**
Considered using "disabled" and using css to style it as if it was enabled but hiding the corresponding box. The problem with this is that `disabled` has ripple effects, and it's not what i'm looking for.

**Additional context**
I imagine it would look like this:
Screen Shot 2019-11-19 at 11 38 11

Bulletninja:
fixes https://github.com/dowjones/react-dropdown-tree-select/issues/306

## What does it do?

Adds a `selectable` prop to avoid rendering checkbox when not needed without disabling it.

## Fixes # (issue)

Closes #306

## Type of change

# Checklist:

### DIFF
--- a/src/tree-node/index.js
+++ b/src/tree-node/index.js
@@ -73,6 +73,7 @@ class TreeNode extends PureComponent {
     showPartiallySelected: PropTypes.bool,
     readOnly: PropTypes.bool,
     clientId: PropTypes.string,
+    selectable: PropTypes.bool,
   }
 
   getAriaAttributes = () => {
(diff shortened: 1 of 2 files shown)


mrchief:
@Bulletninja Thanks for sending this. Looks good! A few things to do:

 - Add it to the docs, demos
 - Add/update unit tests
 - Check for aria compliance and emit proper attributes when `selectable=false`

mrchief:
@Bulletninja were you able to work on these changes?

PR #325, opened 2020-01-19
feat: Prefix instance id when custom node id is passed 

### LINKED ISSUE #315: "x" buttons have same id in HTML DOM
**Describe the bug**
"x" buttons have same id in HTML DOM.
The problem happens when every node has **`id`** property, when nodes have .id, "x" buttons stop using DropdownTreeSelect.props.id.

**To Reproduce**
- Visit my [codesandbox](https://codesandbox.io/s/react-dropdown-tree-select-6ose6)
- Inspect the page, find for "button.tag-remove" (all buttons which's  CSS class = "tag-remove")
- **Actual behavior:** You will notice all found buttons have the same id="2_button"

**Expected behavior**
The buttons' ids should be different between different DropdownTreeSelects (like ${DropdownTreeSelect.props.id}_${node.id}_button).

image

image

mrchief:
## What does it do?

When a user defined node id is specified, it is used instead of an autogenerated on. When the same data is used across multiple components on the same page, this results in duplicate HTML Ids for tags, tag remove buttons.

## Fixes # (issue)

Fixes #315

Also fixes an unreported bug where removing tags by backspace doesn't correctly fire `onTagRemove` event.

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)
- [n/a] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [n/a] This change requires a documentation update

# Checklist:

- [n/a] Updated documentation (if applicable)

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -256,8 +256,7 @@ class DropdownTreeSelect extends Component {
       }
       return
     } else if (e.key === 'Backspace' && tags.length && this.searchInput.value.length === 0) {
-      const lastTag = tags.pop()
-      this.onCheckboxChange(lastTag._id, false)
+      this.onTagRemove(tags[tags.length - 1]._id, true)
     } else {
       return
     }
--- a/src/tag/index.js
+++ b/src/tag/index.js
@@ -3,8 +3,6 @@ import React, { PureComponent } from 'react'
 
 import './index.css'
 
-export const getTagId = id => `${id}_tag`
-
 class Tag extends PureComponent {
   static propTypes = {
     id: PropTypes.string.isRequired,
@@ -37,10 +35,9 @@ class Tag extends PureComponent {
   }
 
   render() {
-    const { id, label, labelRemove = 'Remove', readOnly, disabled } = this.props
-
-    const tagId = getTagId(id)
-    const buttonId = `${id}_button`
+    const { label, labelRemove = 'Remove', readOnly, disabled, getDOMId } = this.props
+    const tagId = getDOMId('tag')
+    const buttonId = getDOMId('button')
     const className = ['tag-remove', readOnly && 'readOnly', disabled && 'disabled'].filter(Boolean).join(' ')
     const isDisabled = readOnly || disabled
 
--- a/src/tags/index.js
+++ b/src/tags/index.js
@@ -7,7 +7,7 @@ import './index.css'
 
 const getTags = (tags = [], onDelete, readOnly, disabled, labelRemove) =>
   tags.map(tag => {
-    const { _id, label, tagClassName, dataset } = tag
+    const { _id, label, tagClassName, dataset, getDOMId } = tag
     return (
       <li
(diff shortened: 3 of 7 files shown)


src/index.js
```diff
@@ -256,8 +256,7 @@ class DropdownTreeSelect extends Component {
       }
       return
     } else if (e.key === 'Backspace' && tags.length && this.searchInput.value.length === 0) {
-      const lastTag = tags.pop()
-      this.onCheckboxChange(lastTag._id, false)
+      this.onTagRemove(tags[tags.length - 1]._id, true)
```
mrchief: @ellinge Do you remember why it was done like this? 

The problem with previous approach was that `onTagRemove` was never fired if you remove tags with `backspace` key. And thus `getNextFocusAfterTagDelete` was also not being invoked.

ellinge: Propably because the focus should be kept on the search field after tag removal?
We are checking if they/you are using backspace in an empty search field and there are tags left. In this case remove the last one (and keep focus on the search field). What is the result of getNext in this case? The search field?

mrchief: Hmm, after removing the last tag the focus goes back to input. I’d need to double check but I’m that’s the case as it would have stood out. 

So if we wanna keep focus on the search field after tag removal, this code (the keybordNavigation.focus... handler will be unreachable code as no one else invokes it).

mrchief: Here's how it looks on different browser with this change:

**Edge** (the focus is not as pronounced, cursor stays in the input (but input is not in focus - if you type anything it doesn't regisetr)
xi7rTgSoQU

Chrome (focus is pronounced, cursor doesn't stay in input until you clear all tags)
8F0ZfZoAA1

ellinge: This is the current implementation. The idea is that the focus is kept on the input field after using backspace. Do you know any WCAG-guidelines stating it should behave differently?

I also checked with react-select, it's how they have done as well. Not stating they are the answer.


- Edge

focusedge

- Chrome

focuschrome

- Firefox

focusfirefox

mrchief: Honestly, I haven't looked into WCAG as I trust your research on it. I've no objections either way. The only question, if we keep current implementation is this:

> So if we wanna keep focus on the search field after tag removal, this code (the keybordNavigation.focus... handler will be unreachable code as no one else invokes it).

I just wanna make sure I'm not missing anything and if this is dead code, then we should take it out. Thoughts?

ellinge: getNextFocusAfterTagDelete

It get's called when tabbing on a specific tag and use backspace

mrchief:
@ellinge @Nogias9x Is this good to go?

ellinge:
@mrchief the change on the resolved conversation should be reverted? Or was that another change that was needed?

mrchief:
> @mrchief the change on the resolved conversation should be reverted? Or was that another change that was needed?

Yeah it should be. It's been a while so let me take a deep dive.

PR #338, opened 2020-02-25
feat(client): Makes the lists sortable

wandererfrog:
## What does it do?
It makes the list of tags sortable. It uses react lib `react-sortablejs`

It does not pass the test for strict control over the HTML structure of lists. 
`"message": "List item does not have a <ul>, <ol> or role=\"list\" parent element"`

But for the intended functionality we need to add a wrapper around the tags.

### DIFF
--- a/docs/src/stories/Simple/index.js
+++ b/docs/src/stories/Simple/index.js
@@ -45,6 +45,7 @@ const Simple = () => (
       onNodeToggle={onNodeToggle}
       onFocus={onFocus}
       onBlur={onBlur}
+      sortable
       className="demo"
     />
   </div>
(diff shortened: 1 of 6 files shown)


wandererfrog:
Hum, I thought I was merging into my fork. 
Anyway, it does not pass the test for the list structure. But I believe that you are not interested in DOM manipulations. So you can close this PR.

mrchief:
Well, my take is that sorting can be achieved simply by sorting the data and that allows flexibility as to what sort logic one wants to use. Alphabetic sorting is one of the ways and it may not make sense in every case. One can always cram in more logic into the component but doing so makes it less flexible. It's a tradeoff so we want to be careful about it. Makes sense?

wandererfrog:
Yes, of course. We just needed that functionality ourselves. The user is supposed to order as they see fit. But yes, trade off.

mrchief:
I'm curios though, why use react-sortable instead of returning sorted tags from treeManager.

Issue #337, opened 2020-02-25
Animating expanding and collapsing could be a UX necessity


deniskabana:
**Is your feature request related to a problem? Please describe.**
For `react-dropdown-tree-select` everything is rendered as a single unordered list with padding added on `node.style` (for tree list with expanded children). I suppose this is due to CSS hiding when filtering (which is quite clever and fast) but kills hierarchy (elements are not wrapped). Makes it very very difficult to implement sliding animation for reveal (close animation is not as necessary for our use-case and from a UX standpoint could be omitted or just very quick).

**TL;DR:** Expanding children in a tree results in mounting and dismounting list items to the top-level row. Very hard to add animation to this type of usage.

![ezgif-6-513831ddc4f8](https://user-images.githubusercontent.com/8308865/75254274-6a8ae600-57e0-11ea-9db3-dd49732d31d7.gif)
Sandbox of the setup in GIF:
https://codesandbox.io/s/cool-noether-p3f41?fontsize=14&hidenavigation=1&theme=dark

**Describe the solution you'd like**

1. Wrap every list item content in a div, like `<li class="node tree"><div class="list-item-container">...</div></li>` and wrap nested children in a `ul > li` structure as well. I feel like I am missing something here, though.
2. Add an option to delay unmounting by any config / prop (e.g. `delayUnmount?: number = 0`) and **create new callbacks** -> `nestedChildrenDidMount(nestedChildren: TreeNode[] | string[])` and `nestedChildrenWillUnmount(timeUntilUnmount: number)`. _Doesn't solve revealing animations however._
3. Consider adding callback specifically to let the users decide whether they can live with performance penalty on manipulating huge lists manually. That could be something like just revealing list of DOM nodes in a specific callback.

**Describe alternatives you've considered**
Right now, to achieve the desired result, I am using a hack - linear transition and style/className manipulation. Easing could have severe impact on performance and I don't think adding custom JS easing for multiple elements is worth the added overhead to the code. See https://codesandbox.io/s/cool-noether-p3f41?fontsize=14&hidenavigation=1&theme=dark for code.

**Additional context**
I'd be happy for any suggestions or ideas in this matter.

deniskabana:
## Update 
For anyone interested, I have been working on another solution. This is a drop-in solution, that wraps the `react-dropdown-tree-select` in another component that intercepts `onNodeToggle(currentNode)` callback. This is what it looks like:
[...]

PR #346, opened 2020-03-24
feat: Add new option expandAllAncestors

osmancode:
## What does it do?

Added new option which expands by default the parent nodes of a checked child.

## Type of change

# Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,7 @@ A lightweight and fast control to render a select component that can display hie
     - [simpleSelect](#simpleselect)
     - [radioSelect](#radioselect)
   - [showPartiallySelected](#showpartiallyselected)
+  - [expandAllAncestors](#expandAllAncestors)
   - [showDropdown](#showdropdown)
     - [initial](#initial)
     - [always](#always)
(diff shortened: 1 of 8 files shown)


mrchief:
@osmancode One can do this by setting the expanded prop in the data. And that should be the source of truth. Another reason this is not a prop is to keep the API simple.

osmancode:
I see your point and I agree with you. 
The issue is iterate over the tree an add the expand attribute to each parent has some performance penalty because I have the dataset separated from the selected values.

mrchief:
Well, someone has to do the iteration. My hesitation with building this in the component is due to these reasons:

 - it adds performance penalty in the tree initialization path (it is already a lot complex than what I'd like to have but we need to draw the line somewhere)
 - it adds code complexity because travelling upwards is more complex than descending
 - it adds another prop that can be avoided (I don't want this to become another rc-tree component)
 - we also need to make it play nice other options (like `hierarchical`, `partialSelect` etc) which adds to the maintenance burden

You mentioned that you don't want to bear the penalty of iterating over the data - well, I'm not sure I would want this component to absorb that cost and pass it on to every user. :)

That said, there are a number of ways we can find a middle ground:

 - have the data producer (API or whatever) add the property when it's assembling the dataset
 - make an HoC that does the iteration for you (something like [tree-node-paths](https://dowjones.github.io/react-dropdown-tree-select/#/story/tree-node-paths-hoc))
 - or something else that you can think of?

I've said this [before](https://github.com/dowjones/react-dropdown-tree-select/issues/337#issuecomment-592014820) that I'm inclined to support baking in those HoCs into the component as long as they are treeshakable and do not add to the cost of the bundle.

Is that something that would work for you? If so, and if you're interested in sending a PR, I would be glad to work with you on that. If not, then I think you may have to create your own fork for this. 

I'm being upfront with you but this is not an end of discussion. I can be persuaded with good reasoning. :)

osmancode:
Totally agree, thanks for the ideas. :)

PR #370, opened 2020-05-06
feat: Focus first match on search

MJRuskin:
## What does it do?

When searching, if the current focus is not a match, it will focus the first match.

## Fixes # (issue)

No issueNumber

## Type of change

# Checklist:

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -142,17 +142,25 @@ class DropdownTreeSelect extends Component {
   }
 
   onInputChange = value => {
-    const { allNodesHidden, tree } = this.treeManager.filterTree(
+    const { allNodesHidden, tree, matches } = this.treeManager.filterTree(
       value,
       this.props.keepTreeOnSearch,
       this.props.keepChildrenOnSearch
     )
     const searchModeOn = value.length > 0
 
+    const { currentFocus } = this.state
+    if (!matches.includes(currentFocus)) {
+      const currentFocusNode = currentFocus && this.treeManager.getNodeById(currentFocus)
+      const firstMatchNode = this.treeManager.getNodeById(matches[0])
+      keyboardNavigation.adjustFocusedProps(currentFocusNode, firstMatchNode)
+    }
+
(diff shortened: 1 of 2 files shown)


MJRuskin:
Just realised this PR changes how some of the navigation tests would work - since the first match is automatically focused before navigating with the arrow keys.

For fixing the tests, would you prefer me to: 
- "correct" the original tests and create an additional test to demonstrate the new feature
or
- integrate a check for this new feature into the original two tests and correct the outcome

(I am assuming the second option is more preferable)

mrchief:
Hey! Thanks, for sending this.

> When searching, if the current focus is not a match, it will focus the first match.

So the reason we do not focus anything right now is to let the user keep on searching. It'd be annoying if I'm typing something and I pause for a sec and then focus is lost from the input. 

I'm not sure automatically switching focus is a good idea in that case but I'm happy to hear what your thoughts are.

MJRuskin:
The search input will never lose focus. This change affects which item in the tree is highlighted.

My intention for the change was to allow users to type and then simply press the Enter key to quickly select the first match.

In the current state this requires the additional press of an arrow key to snap to the first match.

I can gladly make this an optional feature instead. I just felt the current implementation was inconsistent when it came to navigation - especially when going back and forth between searching and the normal view.

mrchief:
OK, I see. Yeah, it seems like this will be a good addition. I need to play with it a bit to see it in action but would love to hear what @ellinge thinks as well.

mrchief:
Can we also add some unit tests for this?

MJRuskin:
Just added some tests and fixed the ones that were affected by this change.

I have never used enzyme before, so please let me know if something needs fixing.

marceleq27:
Hi, any updates in this PR? I was searching in docs for this feature and I came here :)

mrchief:
@marceleq27 haven't had time to dig into this. Will try to get it this weekend.

PR #371, opened 2020-05-06
feat: Default hidden nodes stay hidden after search

### LINKED ISSUE #366: How to skip keyboard navigation for hide node of tree?
Is there a way can skip navigation for hide node of tree? Thanks!

https://codesandbox.io/s/naughty-pine-014m1

MJRuskin:
## What does it do?

Allows nodes to be hidden by default and will stay hidden after you have finished searching.
(Requires the fix from #377 to function correctly)

## Fixes # (issue)

Closes #366 
(My understanding is that this would allow for the "fake" items described in this issue)

## Type of change

# Checklist:


MJRuskin:
Please let me know if you would prefer the defaultHidden property to be worked into the flattenTree function

ellinge:
What does this cover for need? Are the node always hidden? Why would you need hidden nodes? You can use https://github.com/dowjones/react-dropdown-tree-select#searchpredicate if you want to exclude nodes as hits based on a custom prop on the node data.

mrchief:
I'm not sure if I fully understand the use case but this is something you can accomplish with tacking on CSS classes on the nodes to be hidden.

MJRuskin:
If this is done via CSS classes, I believe the nodes will still be selectable while hidden if you navigate to them using the arrow keys.

My intended use case would be hidden nodes that only show up when searched for.

mrchief:
> If this is done via CSS classes, I believe the nodes will still be selectable while hidden if you navigate to them using the arrow keys.

Shouldn't be the case if you use `display:none`. To show them while searching, you can then toggle them via CSS as well:

```
.hiddenNode { display: none; }
.searchModeOn .hiddenNode { display: inline|inline-block|block|etc.; }
```

MJRuskin:
I put this together quickly:
https://codesandbox.io/s/hungry-lederberg-t8uc7

The nodes are hidden, but you can still navigate to them using the arrow keys.
getVisibleNodes checks whether a node is hidden with the hide property, so they will always be "visible" in the tree even if they're not displayed.

isuscbrmid:
@MJRuskin how the wip is going on? did you need help to finish this one?.. I tested and the current solution solve the hide problem avoiding reseting the this.tree values all to hide = false. Let me know if you need help to go foward on this one

MJRuskin:
Hi @isuscbrmid thanks for the offer! I think any changes needed would be quite minor and I can finish them up depending on how @mrchief feels I should tackle this.

First we need to get #377 merged, otherwise the hidden nodes won't remain hidden (depending on your use-case). 
Then, if needed, I will work the defaultHidden property into the flattenTree function.

mrchief:
Taking second look at this and I was wondering - how do nodes get marked hidden by default? And if the nodes are hidden and stay hidden, why can't we eliminate them from the `data` altogether?

isuscbrmid:
Because if it is not in the data we cannot set that value (Even if it remain hidden). Currently on Financial Crime product we have that behavior, we dont show the exclude all and "selected all" con the dropdown content, but they are needed on the data prop to can be abble to "select manually it" and show on the input section

mrchief:
@isuscbrmid  If it is not shown, how does the user select it?

isuscbrmid:
let me show.. here we have 2 "hidden nodes" (All and exclude all) they are only selected by the external functionality. User cannot "click" the nodes inside the dropdown content. (it is set when using data props..
Screen Shot 2020-06-02 at 11 14 37

mrchief:
> let me show.. here we have 2 "hidden nodes" (All and exclude all) they are only selected by the external functionality. User cannot "click" the nodes inside the dropdown content. (it is set when using data props..

If they are external, why can't you just manipulate the data accordingly? Why involve the component at all?

mrchief:
Something like this:

`External Select All > update data to set checked true > set component data`
`External Exclude All > update data to set checked false > set component data`

isuscbrmid:
Everything works fine, the only problem is how the keyboard works, it use the hidden props to know if we are able to navigate through the lists. The problem is even if we hide and control it well, since it does not have the hide property you can navigate through the list even if you are not able to see the (all, exclude all). But we cannot set the hidde props, since it is reset when closing the dropdown content

isuscbrmid:
> Something like this:
> 
> `External Select All > update data to set checked true > set component data`
> `External Exclude All > update data to set checked false > set component data`

yes I do something like this without problem, the only problem is how to hide it when using the keyboard navigation stuff

mrchief:
> yes I do something like this without problem, the only problem is how to hide it when using the keyboard navigation stuff

There is nothing to hide in this case. The Select/Exclude are external components, aren't they?

isuscbrmid:
they are part of the tree data, since we need them to be selected in some cases by the external code, and not the user "click". We need them to show them on the input section. I think the problem is on how Keyboard navigation works, looking into the hidden props and ignoring if they have a display none or something.

isuscbrmid:
On my screenshot I have "Select all" and "Exclude all", the user cannot interact with them. These values are choosen for a external behavior that modify the data provided to the component.

mrchief:
@MJRuskin Do you still need this? If so, then I think we need to talk about your use case. 

Adding a hidden node that always stays hidden is a redundant feature - it's a no-op basically.  

From what I gather, people are using these hidden nodes to accomplish things that are not directly supported by the component. In other words, this feels like a hack or band-aid. I'd much rather fix the real issue.

mrchief:
@all-contributors Please add @MJRuskin for code

@all-contributors Please add @isuscbrmid for code

mrchief:
@all-contributors Please add @isuscbrmid for code

mrchief:
@all-contributors Please add @MJRuskin for code

MJRuskin:
@mrchief Apologies for the lack of input on the comments section.

The hidden nodes don't technically always stay hidden - they only appear when searched.
My use case is for additional options that don't show when looking through the list, but can be searched for and then selected.


Additionally, I believe it would close #366. By adding "fake" nodes named "", which cannot be searched for, you could then select all the (visible) children without selecting the parent.

mrchief:
> The hidden nodes don't technically always stay hidden - they only appear when searched.
> My use case is for additional options that don't show when looking through the list, but can be searched for and then selected.

Would you be able to share screenshots or gifs? I'm curios to see them :)

> By adding "fake" nodes named ""

I'm sorry but I still see the point of these nodes.

PR #374, opened 2020-05-25
fix: Delayed focused update when receiving data changes

isuscbrmid:
## What does it do?
When using controlled data, if the data changes the focused node will be the latest one (the one before, first one is "null").

## Fixes # (issue)
Take the latest state moving the get Focused functionality inside the setState, this ensure to use the latest one.

## Type of change

# Checklist:

### DIFF
--- a/src/index.test.js
+++ b/src/index.test.js
@@ -318,30 +318,19 @@ test('appends selected tags to aria-labelledby with text label', t => {
 })
 
 test('select correct focused node when using external state data container', t => {
-  class DropdownContainer extends React.Component {
-    state = {
-      data: [
-        {
-          label: 'All data',
-          value: '0',
-        },
-        {
-          label: 'iWay',
-          value: '1',
-        },
-      ],
-    }
-    handleChange = ({ value, checked }) => {
-      this.setState(prevState => {
-        const { data } = prevState
-        data[value].checked = checked
-        return { data }
-      })
-    }
-    render() {
-      return <DropdownTreeSelect id={dropdownId} data={this.state.data} onChange={this.handleChange} />
-    }
-  }
+  let data = [
+    {
+      label: 'All data',
+      value: '0',
+      checked: false,
+    },
+    {
+      label: 'iWay',
+      value: '1',
+      checked: false,
+    },
+  ]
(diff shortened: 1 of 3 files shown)


mrchief:
Looks good, can we please add a test for this behavior?

isuscbrmid:
> Looks good, can we please add a test for this behavior?

sure will try to add a integration test where we can show the behavior after clicking and updating an external data state

src/index.test.js
```diff
@@ -316,3 +316,50 @@ test('appends selected tags to aria-labelledby with text label', t => {
   t.deepEqual(wrapper.find('.dropdown-trigger').prop('aria-labelledby'), 'rdts_trigger rdts-0_tag')
   t.deepEqual(wrapper.find('.dropdown-trigger').prop('aria-label'), 'hello world')
 })
+
+test('select correct focused node when using external state data container', t => {
+  class DropdownContainer extends React.Component {
```
mrchief, line 321: We don't need to go this far. The `setState` actions can be simulated using just the component (we do this in other tests)

src/index.test.js
```diff
@@ -316,3 +316,50 @@ test('appends selected tags to aria-labelledby with text label', t => {
...
+    label: 'All data',
+    value: '0',
+    children: undefined,
+    actions: [action],
+  }
+  const wrapper = mount(<DropdownContainer />)
+  const DropdownTreeSelectInstance = wrapper.find('DropdownTreeSelect')
+  DropdownTreeSelectInstance.instance().onCheckboxChange(nodeAllData._id, true)
+  t.deepEqual(DropdownTreeSelectInstance.state().currentFocus, nodeAllData._id)
+  t.deepEqual(DropdownTreeSelectInstance.state().tree.get(nodeAllData._id), {
```
mrchief, line 357: If we're checking for focused prop, then we should make that a specific check instead of comparing the whole state, no? I think that will convey intent better. Thoughts?

isuscbrmid: Sure removing and fixing the test

src/index.test.js
isuscbrmid: the outside setState? since is a external data change? you right will the change in other way

PR #375, opened 2020-05-25
feat: Add prop to skip navigation on nodes

isuscbrmid:
## What does it do?

Currently if I want to hide a node, I can do this with css, but the keyboard feature is able to reach those nodes. We cannot use the "hide" prop in each node, since it is overwritten when "restoringDefaults".

## Fixes # (issue)

Add a custom/optional prop to the nodes to disable keyboard navigation on them.

## Type of change

# Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -270,6 +270,7 @@ Data for rendering the tree select items. The object requires the following stru
   expanded,       // optional: If true, the node is expanded (children of children nodes are not expanded by default unless children nodes also have expanded: true).
   className,      // optional: Additional css class for the node. This is helpful to style the nodes your way
   tagClassName,   // optional: Css class for the corresponding tag. Use this to add custom style the pill corresponding to the node.
+  disableNavitaion, //optional: Allows to slip the keyboard navigation on this node.
   actions,        // optional: An array of extra action on the node (such as displaying an info icon or any custom icons/elements)
   dataset,        // optional: Allows data-* attributes to be set on the node and tag elements
   isDefaultValue, // optional: Indicate if a node is a default value. When true, the dropdown will automatically select the node(s) when there is no other selected node. Can be used on more than one node.
--- a/src/index.keyboardNav.test.js
+++ b/src/index.keyboardNav.test.js
@@ -52,6 +52,27 @@ test('other key presses does not open dropdown on keyboardNavigation', t => {
   })
 })
 
+test('should skip navigation on a node with disableNavigation = true', t => {
+  const tree = [
+    {
+      label: 'All data',
+      value: '0',
+      disableNavigation: true,
+    },
+    {
+      label: 'iWay',
+      value: '1',
+    },
+    {
+      label: 'KDB',
+      value: '2',
+    },
+  ]
+  const wrapper = mount(<DropdownTreeSelect data={tree} />)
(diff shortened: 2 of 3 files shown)


mrchief:
I don't think we need a new prop for this. Skipping all keyboard actions on hidden nodes can be the default behavior.

@ellinge Any apparent concerns if we do this?

mrchief:
Also, please add tests for these changes.

isuscbrmid:
> Also, please add tests for these changes.

no sure if I understand, more tests using that prop?

isuscbrmid:
> I don't think we need a new prop for this. Skipping all keyboard actions on hidden nodes can be the default behavior.
> 
> @ellinge Any apparent concerns if we do this?

we need that because on https://github.com/dowjones/react-dropdown-tree-select/blob/02491b516efc00aa84e377e4ed87b7ac6acfe786/src/tree-manager/index.js#L125 we set the wholes nodes to false. So "hidden prop" will not work

MJRuskin:
@isuscbrmid would you be ok with something like this #371 ?

Any nodes that were initially hidden in the data will remain hidden (and skipped by the keyboard navigation).
They only show up during search.

Unless of course you want a completely hidden "fake" node, and then you can label it "" which makes it is also unsearchable.

mrchief:
> We cannot use the "hide" prop in each node, since it is overwritten when "restoringDefaults".

Not sure I understand this bit. 

In general, I'm usually hesitant to add more props like these ones as it increases the maintenance area in general while it provides little value to only a handful of edge cases. 

You also have to analyse the impact of this change from an accessibility standpoint. 

Also, can we have a CodeSandbox to reproduce the problem?

isuscbrmid:
Sure i will send you the sandbox tomorowEl 25 may. 2020 9:37 p. m., Hrusikesh Panda <notifications@github.com> escribió:

We cannot use the "hide" prop in each node, since it is overwritten when "restoringDefaults".

Not sure I understand this bit.
In general, I'm usually hesitant to add more props like these ones as it increases the maintenance area in general while it provides little value to only a handful of edge cases.
You also have to analyse the impact of this change from an accessibility standpoint.
Also, can we have a CodeSandbox to reproduce the problem?

isuscbrmid:
> #371

yes it seems it can be the solution, I will double check if this one work, will close my Pr

mrchief:
> We cannot use the "hide" prop in each node, since it is overwritten when "restoringDefaults".

I guess I'm confused because there is no such prop; or a method called `restoringDefaults`.

isuscbrmid:
sorry i write down wrongly the name, here the hidde is set to false, even if we pass a hidde = true at the beginning.. So only in the beginning it will hidde the node, but after closing the dropdown this method is triggered removing the hidde = true
https://github.com/dowjones/react-dropdown-tree-select/blob/02491b516efc00aa84e377e4ed87b7ac6acfe786/src/tree-manager/index.js#L123

isuscbrmid:
I wil test @MJRuskin  PR I thing the way he did avoid that overwrite can help me

isuscbrmid:
@mrchief we can close this.. The solution provided by @MJRuskin will work It avoid overwriting the default nodes hide prop when restoring nodes.

PR #377, opened 2020-05-28
fix: Correctly hide non-matching parents on search

MJRuskin:
## What does it do?

The "_isMatch" property is never cleared, so non-matching parents will still be shown (not greyed-out) and can be focused (i.e. navigation with the arrow keys)..

This fixes that issue by setting all parents to hidden, unless they themselves are a match. They will still be shown with keepTreeOnSearch, because the "matchInChildren" property is set to true.

## Fixes # (issue)

No issueNumber

## Type of change

# Checklist:


mrchief:
Looks good! Can we please add a failing test that would fail without your fix? I want to make sure any future refactorings don't re-introduce this bug.

MJRuskin:
Sure thing, I'll look into how to reproduce the bug in a test.

MJRuskin:
@mrchief looking into the keyboardNavigation tests, it appears that some of the ones that work might be written incorrectly. However, this depends on what author's original intention was, so I just wanted to get some clarification.

For example, the test _"can navigate with keepTreOnSearch on keyboardNavigation"_, the ArrowDown is pressed three times.
This _looks_ like it is meant to navigate from "ccc 1" -> "ccc 3" -> "bbb 1"
However, this isn't something you should be able to do, and what actually happens is that it navigates from "bbb 1" -> "bbb 2" -> "bbb 1"

ellinge:
> However, this isn't something you should be able to do, and what actually happens is that it navigates from "bbb 1" -> "bbb 2" -> "bbb 1"

The result set is filtered:
`wrapper.instance().onInputChange('bb')`

MJRuskin:
Hi @ellinge, I am aware that the result set is filtered. I just wanted to confirm whether this was your intention when writing the test.
i.e. Did you want it to cycle through these two options?

ellinge:
Yes, it’s a cycle to ensure only the filtered selectable nodes (bbb) are passed even if the parent is visible (the second check). Maybe was a copy of another navigation test but still fills a purpose.

MJRuskin:
Thanks @ellinge! That has helped me a bit with understanding the existing tests.

I think I have a test working now for this bugfix. I will clean it up and push soon.

ellinge:
The discussed test becomes obselete by your added more fine grained test so that could be removed (the old one).

MJRuskin:
I can combine them together (so it keeps the cycling of filtered options)

ellinge:
Well your solution of checking initial focus keeps the confusion away of which path it took to bbb 2?

PR #378, opened 2020-06-04
fix: action click node focus

isuscbrmid:
## What does it do?
When clicking a node on the action section, it is not focused or still focuses on the previous one. This is confusing.

## What this PR do?
Update the tree manager with the current node where the action was clicked with the _focused correct value.

## Type of change

# Checklist:

Note: I had to update one of the tests since the returned node on the callback has the _focused value.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -212,7 +212,17 @@ class DropdownTreeSelect extends Component {
   }
 
   onAction = (nodeId, action) => {
-    this.props.onAction(this.treeManager.getNodeById(nodeId), action)
+    const { currentFocus } = this.state
+    const { searchModeOn } = this.props
+    const currentFocusNode = currentFocus && this.treeManager.getNodeById(currentFocus)
+    const node = this.treeManager.getNodeById(nodeId)
+    const tree = searchModeOn ? this.treeManager.matchTree : this.treeManager.tree
+    keyboardNavigation.adjustFocusedProps(currentFocusNode, node)
+    this.setState({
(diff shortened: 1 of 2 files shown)


mrchief:
Will you able able to paste a gif or screenshots? I want to see this "awkwardness". Action items were never meant to focus the nodes. You'd place the focus on the action item itself if you think about web accessibility standards since you're clicking a button/link (action item).

isuscbrmid:
On my keyboard version, we focused the checkbox, and not the whole node. 

Since it is focusing on the whole node lists. It can be confusing to not see the node "focused" when I do some action there.

isuscbrmid:
action-button

isuscbrmid:
Screen Shot 2020-06-08 at 12 00 36
This is "focused", don't know why on the gif is not colored.

mrchief:
You can add a hover or active CSS to achieve the highlight, no?

isuscbrmid:
Yeah, but how we can remove that highlight, it will affect the styles when using keyboard actions. I think a good approach is to focus the node internally since if I start to use the keyboard you can start from the latest nodes with an interaction.

mrchief:
Keyboard should be already highlighting items as you navigate. Actions is not the same as selecting a node. But I want to hear what others think. I'm kinda on the fence on this one. 

@ellinge @Gregcop1 @MJRuskin thoughts?

mrchief:
not stale

PR #379, opened 2020-06-04
feat: add controlled prop

isuscbrmid:
## What does it do? (Help needed label required)
Based on the https://codesandbox.io/s/select-all-checkbox-with-react-dropdown-tree-select-d0m00?file=/DropdownHOC.js, we see that when changing the value to "checked: false" through the props, sometimes is not updated.

## What did I found?
Since the node is a pure Componente is not updated in a specific scenario. If we change this to a normal component if will affect the performance. 
Another solution could be to ensure that the component is upgraded to checked = true, and then the data change will update it to checked = false.

## What this PR do?
It is just an idea, actually, I need help on which direction I could try. The idea works but since we don't want to add more props/things to the component I need to find another way to ensure the updates inside the component when it relies on external data changes.


mrchief:
Hey, yeah we should discuss this more. The component should update when things change and you shouldn't be needing to add props to make it do what it is supposed to do in the first place.

I looked at your sandbox and I was wondering if it happens outside of the HoC? If so, then it's a bug. If not, then I thing the HoC is where the issue lies.

mrchief:
I checked that switching to Component doesn't change the behavior. Furthermore, the `checkbox` is a controlled component. 

For this particular use case, it seems React or DOM was skipping UI updates given the nature of the HoC. The component was updating correctly as can be seen by closing and opening the dropdown again. 

The solution was to `requestAnimationFrame(this.setState(...), 0)` so that UI updates are not skipped.

PR #382, opened 2020-06-11
fix: Add default onAction prop

### LINKED ISSUE #380: TypeError: n.props.onAction is not a function
**Describe the bug**
onAction property is optional based on PropTypes and typescript types, however it's being called in the onClick handler, that assume onAction property is a function.

**To Reproduce**

Create data with nodes, that have actions array populated, 
click on the node with action.

Note it was working fine on original fork version of sandbox (lib v1.13.0)
CodeSandbox Link: https://codesandbox.io/s/react-dropdown-tree-select-bvlff?file=/index.js

**Expected behavior**
Clicking on the node, should not throw an error.

**Screenshots**
image

**Environment*
 - Browser Chrome 83
 - OS: Linux

kYem:
## What does it do?

PropTypes define onAction property as optional, however in class its
used unconditionally. Trying to click the action without passing onAction handler cause type error.

Fixes #380 

## Type of change

# Checklist:


src/index.test.js
```diff
@@ -107,6 +107,13 @@ test('notifies on action', t => {
   t.true(handler.calledWithExactly(node0, action))
 })
 
+test('notifies on action without onAction handler', t => {
+  const { tree } = t.context
+  const wrapper = mount(<DropdownTreeSelect id={dropdownId} data={tree} showDropdown="initial" />)
+  wrapper.find('i.fa-ban').simulate('click')
+  t.pass()
```
mrchief, line 114: `t.pass`?

mrchief:
@all-contributors Please add @kYem for bug, code

src/index.test.js
kYem: I'm not entirely sure what I need to check here :thinking: 
The test is only checking if the click event does not trigger error, therefore it pass.

Any suggestions what would be correct test for it otherwise?

mrchief: Well, there is literally no checks/assertions in this test. It adds no value. I'm not sure if there's a way to check for those type of warnings during a test run. If we can't find a way to ensure that a specific warning is not emitted, then I'd rather not have this test.

mrchief: I think one way we could check for this is spy console.warn or console.error (whichever is emitted by React) and then check that it doesn't get called. 

You can remove your default prop and this test should fail - which would confirm that this approach works :)

kYem: Well, because it's actual JavaScript TypeError, nothing to do with React itself, it already fails by default without any checks, if default prop is removed, like this:

```javascipt
index › notifies on action without onAction handler

....../javascript/react-dropdown-tree-select/src/index.js:216

   215:   onAction = (nodeId, action) => {                                   
   216:     this.props.onAction(this.treeManager.getNodeById(nodeId), action)
   217:   }                                                                  

  Error thrown in test:

  TypeError {
    message: '_this.props.onAction is not a function',
  }

```

I changed from pass() to spy for console error, but honestly I don't see any value in it.

Let me know if this acceptable now

mrchief: Well `pass` is of no value. This test actually checks and will warn in future if we break it again. I'll check it and merge (mostly looks good).

kYem:
@mrchief Is this going to get merged?

PR #410, opened 2020-08-06
fix: clear search input on resetting state

### LINKED ISSUE #409: Search input is not cleared on selection
When using both `clearSearchOnChange` and `inlineSearchInput` options, the search input it not cleared inside the opened dropdown upon selecting one of the search results. This is a weird UX — I think the input should be cleared on selection too.

I found [this piece of code](https://github.com/dowjones/react-dropdown-tree-select/blob/develop/src/index.js#L93) which is not clear for me..

https://www.loom.com/share/ae3bef73fd7345b78468dd853955973c

berdyshev:
When using both `clearSearchOnChange` and `inlineSearchInput` options, the search input is not cleared inside the opened dropdown upon selecting one of the search results. 

We need to check whether the input ref is accessible but not the case when `inlineSearchInput` is not set.

Fixes #409


berdyshev:
@mrchief please check

mrchief:
Will check and let you know.

mrchief:
@all-contributors Please add @berdyshev for code,bug

mrchief:
@all-contributors Please add @berdyshev for code and bug

PR #317, opened 2019-12-19
feat: Added tagLabel option to make node and tag labels differ

Bulletninja:
## What does it do?

Adds a tagLabel prop for those use cases where you want to display something different in a tag label compared to the checkbox one.

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# Checklist:

- [ x ] I have performed a self-review of my own code
- [ x ] I have commented my code, particularly in hard-to-understand areas
- [ x ] Updated documentation (if applicable)
- [ ] Added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ x ] My changes generate no new warnings

PR #412, opened 2020-08-07
feat: Add support to use different labels for tag and node

### LINKED ISSUE #316: Different label in tree vs Tag
Currently i have this use case where some leafs have the label, and it's alright inside the tree, since the tree itself gives it enough context to distinguish the entries, The problem is how to distinguish them in the tags:

Screen Shot 2019-12-18 at 11 19 53

I would like to be able to specify a label specific for the tag, like some sort of expanded version of the label.

I added a field called `tagLabel` with the name i want for the Tag, and then i assumed i could do this simply by extending the `onChange` event via the `HOC pattern`. And, fair enough i get what i want in the Tag label, but then the original label also changes:

Screen Shot 2019-12-18 at 11 30 35

Is it currently posible to do this?

m4theushw:
## What does it do?

Adds a `tagLabel` prop for those use cases where you want to display something different in a tag label compared to the checkbox one. I reused the work done in #317.

## Fixes # (issue)

Fixes #316 

## Type of change

# Checklist:


mrchief:
Thanks for sending this @m4theushw!

@all-contributors please add  @m4theushw for code

PR #425, opened 2020-09-16
fix: Trigger onFocus/onBlur

### LINKED ISSUE #418: onBlur has inconsistent behavior.
**Describe the bug**
On Blur doesn't get triggered on focus out.
When a selected option is removed with the 'x'button onBlur isn't triggered, but when it is removed from the dropdown, on blur is triggered. 

**To Reproduce**
CodeSandbox [here](https://codesandbox.io/s/using-bootstrap-styles-with-react-dropdown-tree-select-forked-lpe2t?file=/index.js) with a console.log attached to onBlur. 
 - focus in and focus out the select, on blur doesn't get triggered. (navigating with tab)
 - remove a selected option from the control (using the x button) on blur doesn't get triggered, but is you uncheck a selected option, onBlur gets triggered. 

**Expected behavior**
onBlur to be triggered on focus out.
onBlur to be triggered when removing a component with the x button 

 - Browser [e.g. chrome v84, firefox v78]

lmotioc:
## What does it do?

Updates onFocus and onBlur events for the whole component to act like a composite input. 
Triggers onFocus on first input focus of the select component and onBlur on last input blur of the select component.

## Type of change

# Checklist:

- [ x] I have performed a self-review of my own code
- [ x] I have commented my code, particularly in hard-to-understand areas
- [ n/a] Updated documentation (if applicable)
- [ n/a] Added tests that prove my fix is effective or that my feature works
- [ x] New and existing unit tests pass locally with my changes
- [ x] My changes generate no new warnings


AdrianParasciuc (dismissed):

lmotioc:
i'm not sure i should refactor the main component to fix the codeclimate issue. any advice @mrchief  sir? 😁

src/index.js
```diff
@@ -106,40 +107,59 @@ class DropdownTreeSelect extends Component {
...
   }
 
   componentWillReceiveProps(nextProps) {
     this.initNewProps(nextProps)
   }
 
+  onBlur = () => {
+    // setTimeout runs afterwards in the event.
+    // If onFocus is triggered immediately in a child component, clearTimeout will stop setTimeout to run.
+    this._timeoutID = setTimeout(() => {
```
mrchief, line 120: setTimeout/clearTimeout seems like a fragile solution. Not just from a timing perspective but I think it makes it harder to move towards hooks. I was going to ask if you have you looked at using requestAnimationFrame in lieu of setTimeout but then I see that it's being used in conjunction with clearTimeout so RAF might not be an option.

I'm curious though - why take on the additional burden of keeping track of things instead of detecting outside clicks?

src/index.js
```diff
@@ -157,12 +177,15 @@ class DropdownTreeSelect extends Component {
...
     const { tags: prevTags } = this.state
     this.onCheckboxChange(id, false, tags => {
-      if (!isKeyboardEvent) return
-
-      keyboardNavigation.getNextFocusAfterTagDelete(id, prevTags, tags, this.searchInput).focus()
+      const nextFocus = keyboardNavigation.getNextFocusAfterTagDelete(id, prevTags, tags, this.searchInput)
+      if (nextFocus) {
+        nextFocus.focus()
+      } else {
+        this.onBlur()
```
mrchief, line 187: This may not be an onBlur as the user would still be in the input, wouldn't they?

types/react-dropdown-tree-select.d.ts
```diff
@@ -114,6 +114,7 @@ declare module 'react-dropdown-tree-select' {
     searchInput: HTMLInputElement
     keepDropdownActive: boolean
     handleClick(): void
+    _timeoutID: number
```
mrchief, line 117: Why do we need to expose this?

src/index.js
lmotioc: I agree it looks a bit like smelly code, but it does not delay in terms of time but on one cycle on the event loop. 

I choose this solution, first, because it builds over the existing onFocus/blur events from the dom and it doesn't need to treat corner cases. in my opinion this is easier to maintain then associating the blur focus with other events like opening and closing the dropdown. Secondly,  the show stopper for me was the complexity of having the blur triggered when using keyboard navigation in a form. you could be listening on tab on the last element but you can not be sure which one is it,you have to check the props. it seemed to add to much complexity on the code.

i also tried attaching a handler like it is done for closing the dropdown when clicking outside but i ran into some corner cases, and i am sure i cannot think of all of them. also it might be harder to maintain.

src/index.js
lmotioc: i can not see a case where the component is on focus. if the focused element is removed, and there is nothing else you can focus on, doesn't that mean the component loses focus? what am i missing?

PR #427, opened 2020-09-16
feat: Always show tree in ".dropdown-content"

wchen-obj:
## What does it do?

This change will always show the tree HTML in ".dropdown-content" when search in input so custom scrollbar package "Simplebar" can handle the list properly.
Please check the [Codesandbox](https://codesandbox.io/s/modern-sun-4vtfw) for a detail visual review of original package.

- Open the dropdown and check the Simplebar custom scrollbar is applied to ".dropdown-content > ul" (the custom scrollbar shows always).
- Search "xxx" in search input and "No matches found" shows.
- Clear the search and the custom scrollbar is gone.

## Type of change

# Checklist:


mrchief:
@wchen-obj Thanks for sending this. The Codesandbox uses the latest release and not your version so I'm not sure if this is not a breaking change. From what I can tell, with your change, the tree would be visible when no matches are found which is not the expected functionality.

wchen-obj:
With this change the tree's wrapper HTML will not being removed and while search the list items removes.
```
<div class="dropdown-content">
  <span class="no-matches">No matches found</span>
  <ul class="root searchModeOn" role="tree" aria-multiselectable="true">
    <div>
      <div class="infinite-scroll-component " style="height: auto; overflow: auto;"></div>
    </div>
  </ul>
</div>
```

wchen-obj:
> @wchen-obj Thanks for sending this. The Codesandbox uses the latest release and not your version so I'm not sure if this is not a breaking change. From what I can tell, with your change, the tree would be visible when no matches are found which is not the expected functionality.

Sorry I don't know how to add my branch as a dependency in codesandbox could you please help with this so I can update it, thank you very much.

This codesandbox just show the issue I ran into when I use a custom scrollbar package to try to style tree dropdown.

mrchief:
> Sorry I don't know how to add my branch as a dependency in codesandbox could you please help with this so I can update it, thank you very much.

npm supports pointing to any git pointer in the package json. I haven't tried it but Codesandbox might also support it. 

> This codesandbox just show the issue I ran into when I use a custom scrollbar package to try to style tree dropdown.

I'm not sure what I supposed to expect so apologies for not getting it right away. This issue has been brought up several times in the past and I'm hesitant to bake in a custom scrollbar.

PR #429, opened 2020-09-20
Sync hooks branch with develop

m4theushw:
## What does it do?

I pulled changes from the `develop` branch so we can continue implementing the Hooks-based version.


mrchief:
Many thanks @m4theushw !

PR #439, opened 2020-11-03
feat: Set the tabIndex on tree select

marianjuhas:
## What does it do?

Tab index can be set for tree select 

## Type of change

- [x] New feature (non-breaking change which adds )

# Checklist:


mrchief (approved):
Looks good! Thanks for sending this

PR #448, opened 2020-11-27
fix: Issue with searchPredicate function second argument(searchTerm)

manish-dalal:
## What does it do?

An issue with searchPredicate function second argument(searchTerm). Always received lowercase of search string. Needed the actual search string that is typed by a user.
Please check the issue on given eg:- [react-dropdown-tree-select-searchpredicate-issue](https://codesandbox.io/s/react-dropdown-tree-select-searchpredicate-issue-2hwh4
) 

## Fixes # (issue)

Updated position of converting searchTerm to lowercase.

## Type of change

- [ x ] Bug fix (non-breaking change which fixes an issue)

# Checklist:

- [ x  ] I have performed a self-review of my own code
- [ x ] I have commented my code, particularly in hard-to-understand areas

### DIFF
--- a/src/tree-manager/index.js
+++ b/src/tree-manager/index.js
@@ -89,7 +89,7 @@ class TreeManager {
   }
 
   filterTree(searchTerm, keepTreeOnSearch, keepChildrenOnSearch) {
-    const matches = this.getMatches(searchTerm.toLowerCase())
+    const matches = this.getMatches(searchTerm)
 
     const matchTree = new Map()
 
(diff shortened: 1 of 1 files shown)


mrchief:
@manish-dalal thanks for the CodeSandbox and the PR. Trying to understand the actual issue here: are you trying to make sure the "Searched String" label shows the search string as user typed it?

image

Or is it that search is broken somehow?

manish-dalal:
@mrchief thanks for the reply. I just want to implement case sensitive searching using searchPredicate props. On searchPredicate second argument(searchTerm), we are getting the actual string typed by the user. And currently, it is always the lowercase of the actual string. So I need the actual string, not the lowercase of that. For more clarification, I also updated the CodeSandbox example, please check [react-dropdown-tree-select-searchpredicate-issue](https://codesandbox.io/s/react-dropdown-tree-select-searchpredicate-issue-2hwh4
)

mrchief:
So if I understand correctly, when you search for `acc`, `Acc` shouldn't show up and vice versa

image

And that doesn't work as `searchPredicate` receives lowercased version of the search string, correct?
image

manish-dalal:
Correct

PR #482, opened 2021-04-12
fix(compilation): Fixing the searchModeOn compilation problem

### LINKED ISSUE #472: when searching a tree, only the parent of child results show (only when app is hosted)
**Describe the bug**
when using the search feature, only the parent of the matching node appears (the 'HainesAttract' node has a child called 'Technology', see next screenshot)
image

However, when the app is running locally on my machine, everything works as expected - I type 'tech' and the relevant child node 'Technology' (and it's children) appear.
image

**To Reproduce**
the issue occurs only on a hosted environment. 

**Expected behavior**
see above.

**Screenshots**
see above.

 - Browser: 89.0.4389.82 (Official Build) (64-bit)
 - OS: Local: Windows x64, Hosted: Azure Hosted App (Unix 4.15.0.112)
 - node version: 12.16

**Additional context**
using react-dropdown-tree-select version 2.5.1.
It might be that the info I provided is not enough, but I'm not quite sure what to look into at this point, so any questions/requests for further into are welcome.

Kulgar:
## What does it do?

See issue #472 and discussion #479 for details and context. 

Only moved the "shouldRenderNode" method within the Tree component and changed its argument a bit to solve a compilation problem. I actually didn't get why that method was outside of the component... any reasons for that? 
I put it in the component because I thought this might solve the compilation problem I faced (along with the argument change). 

I needed this fix to properly compile this lib with react-scripts. I don't know why the bug I found actually happens, if it is something in the original code or the way react-scripts minify that lib dist files... but the fix I did actually solve my problem so... 

I didn't actually add any new unit tests... I'm not even sure how I could do a proper test for that bug as it only happens at compilation time within an app that actually uses that lib...

Please, let me know if I could improve my PR in any way. 
Thanks!

## Fixes # (issue)

Fixes #472 

## Type of change

# Checklist:

### DIFF
--- a/src/tree/index.js
+++ b/src/tree/index.js
@@ -5,15 +5,6 @@ import React, { Component } from 'react'
 import TreeNode from '../tree-node'
 import { findIndex } from '../utils'
 
-const shouldRenderNode = (node, searchModeOn, data) => {
-  if (searchModeOn || node.expanded) return true
-
-  const parent = node._parent && data.get(node._parent)
-  // if it has a parent, then check parent's state.
-  // otherwise root nodes are always rendered
-  return !parent || parent.expanded
-}
-
 class Tree extends Component {
   static propTypes = {
     data: PropTypes.object,
@@ -68,6 +59,17 @@ class Tree extends Component {
     }
   }
 
+  shouldRenderNode = (node, props) => {
+    const { data, searchModeOn } = props
+    const { expanded, _parent } = node
+    if (searchModeOn || expanded) return true
+
+    const parent = _parent && data.get(_parent)
+    // if it has a parent, then check parent's state.
+    // otherwise root nodes are always rendered
+    return !parent || parent.expanded
+  }
+
   getNodes = props => {
     const {
       data,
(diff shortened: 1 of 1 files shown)


mrchief:
>  I actually didn't get why that method was outside of the component... any reasons for that?

This component started with functional components (what everyone calls hooks these days) and then moved to a class based style. Back in the days react wasn't doing anything special with the functions so to prevent re-renders, I used classes which at least does the shallow check out of the box and allows you to hook into lifecycle events for greater control. This was the only method that could stay **_purely functional_** I believe so I kept it like that. 

Thanks for sending this PR. I'd like to investigate the minification problem a bit more and based on the findings, will merge this PR or fix the root cause.

Kulgar:
Alright, let me know what you find about this minification problem as I'm really interested in understanding the root cause of the bug. 
From my point of view, it seems kinda related to a problem of variable scope, as if, during the minification of the final app, it no longer knows that "searchModeOn" is a variable that comes from the component. That is why I tried to use "props" instead of the variables, which seems to have solved the pb. It's just a guess though.

mrchief:
> From my point of view, it seems kinda related to a problem of variable scope, as if, during the minification of the final app, it no longer knows that "searchModeOn" is a variable that comes from the component. That is why I tried to use "props" instead of the variables, which seems to have solved the pb. It's just a guess though.

Sounds like it's doing some kind of dead code elimination which it shouldn't as the code is not dead! 😄

Kulgar:
Exactly, and it only considers it as dead during the re-minification of the minified file... Really weird ^^'

mrchief (approved):

PR #516, opened 2021-09-09
fix: Adds defensive code to keyboard navigation

mlohmann:
## Fixes # (issue)

Adds defensive code that was causing issues with getNodesMatching not having a tree passed in the arguments. Errors were thrown when calling 'forEach' on the undefined tree. 

## Type of change

# Checklist:

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -263,7 +263,7 @@ class DropdownTreeSelect extends Component {
         })
       }
     } else if (showDropdown && ['Escape', 'Tab'].indexOf(e.key) > -1) {
-      if (mode === 'simpleSelect' && tree.has(currentFocus)) {
+      if (mode === 'simpleSelect' && tree && tree.has(currentFocus)) {
         this.onCheckboxChange(currentFocus, true)
       } else {
         // Triggers close
--- a/src/tree-manager/nodeVisitor.js
+++ b/src/tree-manager/nodeVisitor.js
@@ -11,15 +11,16 @@ const getNodesMatching = (tree, nodePredicate) => {
   const nodes = []
   const visited = {}
 
-  tree.forEach((node, key) => {
-    if (visited[key]) return
+  tree &&
+    tree.forEach((node, key) => {
+      if (visited[key]) return
 
-    if (nodePredicate(node, key, visited)) {
-      nodes.push(node)
-    }
+      if (nodePredicate(node, key, visited)) {
+        nodes.push(node)
+      }
 
-    visited[key] = true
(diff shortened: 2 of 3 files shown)


mrchief:
@mlohmann thanks for sending this. Could you please add a unit test for this as well?

Also, how can I reproduce this?

mlohmann:
The error occurred if you repeatedly hit the navigation arrows when the
filtered options returned were empty. Here's the console output. I've added
a test to the PR.

-Matt

[image: image-2021-08-26-11-51-31-144.jpeg]

mrchief (approved):

mlohmann:
Looks like the PR was approved on the 13th, but seems to be blocked again. Also, do you publish snapshot releases?

mrchief:
@mlohmann seems like the required workflows needed additional approval for first time contributors. I just unblocked it and so once it passes, will do the needful to get it out.

hyewon92cho:
Do you mind to add null check for `!tree.has(prevFocus._id)` as part of this PR?
I'm getting "TypeError: Cannot read property 'has' of undefined" due to this line.
https://github.com/dowjones/react-dropdown-tree-select/blob/master/src/tree-manager/index.js#L283

mrchief:
> Do you mind to add null check for `!tree.has(prevFocus._id)` as part of this PR? I'm getting "TypeError: Cannot read property 'has' of undefined" due to this line. https://github.com/dowjones/react-dropdown-tree-select/blob/master/src/tree-manager/index.js#L283

I don't mind it but in general, I believe prevention is better than cure :). I've a felling `tree` shouldn't be a null by the time the code reaches here so can you help me get to the root of this problem? Is there a codesandbox you can set up? If not, let me know the exact steps to reproduce this.

hyewon92cho:
@mrchief Thank you for the reply! Since this component is used in my company website code so I tried reproducing it in code sandbox with something like below but I couldn't make it😂
  <DropdownTreeSelect
    data={data}
    mode="radioSelect"
    keepTreeOnSearch={true}
    showDropdown="default"
  />
The only thing I know is this component breaks if I filter results by random string and make the available options as 0, and press left or right key. prevFocus was not null nor undefined but tree was undefined when the error occured.

mrchief:
@hyewon92cho I understand your situation. Let's see if we can find a reproducible case. This will help in making sure that we're fixing the right defect and not just applying some band aid. I'd also like to add a unit test for this edge case so that future modifications to the code do not reintroduce the problem.

PR #580, opened 2022-07-13
feat: Do not incur cost of filtering tree if search mode is off

gandhis1:
## What does it do?

When you have tens of thousands of items in a search tree, and you type search input, the cost can be extremely high when you have zero or one characters, especially when you are using a custom search predicate. There is no need to incur the cost of filtering the tree if the text input is zero length. Instead, just reset the search state.

I'll note here that the example in the demo that is supposed to test [a large-scale scenario](https://dowjones.github.io/react-dropdown-tree-select/#/story/large-tree) doesn't appear to be all that taxing because the number of nested nodes is very low. You need a lot of nested nodes to encounter performance issues. In my testing, I have almost 100k nodes at up to 10 levels of nesting. I also have `keepTreeOnSearch` on which probably exacerbates the issue.

This is the first change I would like to do. In a future PR, my intention would be to set the threshold (search term length) at which the search activates. When your search term is "a" you can trigger rendering of thousands of nodes potentially. So this parameter I am envisioning would basically prevent search mode until X characters...with some kind of visual indicator.

## Type of change

# Checklist:

- [X] have commented my code, particularly in hard-to-understand areas
- [X] Updated documentation (if applicable)

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -148,18 +148,21 @@ class DropdownTreeSelect extends Component {
   }
 
   onInputChange = value => {
-    const { allNodesHidden, tree } = this.treeManager.filterTree(
-      value,
-      this.props.keepTreeOnSearch,
-      this.props.keepChildrenOnSearch
-    )
     const searchModeOn = value.length > 0
-
-    this.setState({
-      tree,
-      searchModeOn,
-      allNodesHidden,
-    })
+    if (!searchModeOn) {
+      this.setState(this.resetSearchState())
+    } else {
+      const { allNodesHidden, tree } = this.treeManager.filterTree(
+        value,
+        this.props.keepTreeOnSearch,
+        this.props.keepChildrenOnSearch
+      )
+      this.setState({
+        tree,
+        searchModeOn,
+        allNodesHidden,
+      })
+    }
   }
 
   onTagRemove = (id, isKeyboardEvent) => {


gandhis1:
Tagging @mrchief

gandhis1:
As a temporary step, I ended up forking this package and publishing my own: https://www.npmjs.com/package/@gandhis1/react-dropdown-tree-select

I then tested it in my main application and this fix works exactly as intended. 100k nodes across 10+ levels does not lock up when you have an empty string search term.

mrchief:
@gandhis1 Thanks for sending this! This sounds like a great idea and I'll look into this soon. Been swamped at work lately

gandhis1:
Hi any update on this? It's a pretty tiny PR

mrchief:
Sorry @gandhis1 Will do by end of the week

gandhis1:
Following up, can this be reviewed?

mrchief:
@gandhis1 I'm swamped at the moment but I plan on getting to this, along with the other stuff that's pending. Thanks for waiting.

gandhis1:
Posting to keep this from being stale, PR is awaiting review

gandhis1:
Not stale

mrchief (approved):
Sorry for the delay and thank for sending this

Issue #581, opened 2022-07-15
control the search field


KB-AWARE:
Hi,

I want to know if it is possible to control the search field or have a onInputChange callback?

Thanks.

gandhis1:
So I handled this by passing a custom `searchPredicate` that saves the search term in a `ref`. It's wasteful because that predicate runs on every node, but it at least works.

The part that I do not believe there is an API for is controlling the search field. I previously requested a way to provide an initial input for the search field, and a built-in way to view the current search term, and was told that would be an enhancement.

KB-AWARE:
Hi,

Thanks for you reply.
Finally I used the same trick as you but it is not ideal as the search term return nothing if there is no more items in the filtered options.

The other thing is that it is mandatory that you put options to have a a non null searchTerm. It is annoying as I wanted to fetch the options async.

gandhis1:
Yeah, I have also noticed this doesn't work if `clearSearchOnChange` is selected.

gandhis1:
I have a PR for this, keep open

Issue #576, opened 2022-06-29
Single click to clear search field


bmm-brobbins:
**Is your feature request related to a problem? Please describe.**
After searching, must manually clear search field, via e.g. highlight+delete or backspacing.

**Describe the solution you'd like**
Proposal for an 'x' to appear at right of search field to enable single-click clearing of search contents.

**Describe alternatives you've considered**
Haven't found a solution here or the usual forums.

**Additional context**
Any pointers or code to accomplish this feature in the interim using lib's existing capabilities and/or an HOC would be great.
I'm using inlineSearchInput, btw.  
P.s. Great lib, thank you for publishing it.

mrchief:
@bmm-brobbins I think it's a fair use case but implementing it is bit more than trivial. Apart from the likely UX confusion that extra `x` may create on smaller viewports, we also need to figure out accessibility aspect of it - that includes keyboard nav, tab focus order etc. 

There is something you can do though to mimic the functionality:
 - create your own wrapper control that puts the `x` in the input box (use CSS to position it)
 - on click of this button, you programmatically reset the data (in effect wiping out all selections)

PR #589, opened 2022-08-06
feat: Add new props for search and keyboard navigation

gandhis1:
## What does it do?

- The new prop `searchTerm` can be used to set the initial search term or fully control the search semantics for a higher-order component.
- The new prop `onSearchChange` allows specification of a callback for when the search term changes or the search mode is activated or deactivated.
- The new prop `disableKeyboardNavigation` allows disabling all actions associated with a key down event in the search input box. This is important if you need an input box to behave normally. For instance, when I press SHIFT+HOME, it should select the entire line, so I can press BACKSPACE to clear it.

## Fixes # (issue)

Fixes https://github.com/dowjones/react-dropdown-tree-select/issues/581
Requested https://github.com/dowjones/react-dropdown-tree-select/discussions/463
Fixes https://github.com/dowjones/react-dropdown-tree-select/issues/576

## Type of change

# Checklist:

- [X] Updated documentation (if applicable)


src/index.js
```diff
@@ -50,18 +53,21 @@ class DropdownTreeSelect extends Component {
...
     disablePoppingOnBackspace: PropTypes.bool,
+    disableKeyboardNavigation: PropTypes.bool,
   }
 
   static defaultProps = {
     onAction: () => {},
     onFocus: () => {},
     onBlur: () => {},
     onChange: () => {},
+    onSearchChange: (_) => {},
```
gandhis1, line 64: There is an incompatibility between the `prettier` pre-commit hook and the `codeclimate` checks on GitHub with regard to this line. One tool insists on parentheses being present, the other insists it it on being absent. Is there any way we can use the same linter on both so that one tool does not conflict with the other like this?

mrchief: Hmm, not sure. Will have to check with CC docs to see if they support a lint rc.

mrchief: We can ignore the warning manually in CC dashboard as a last resort

gandhis1:
Bump

types/react-dropdown-tree-select.d.ts
```diff
@@ -103,6 +110,8 @@ declare module 'react-dropdown-tree-select' {
      * search bar, the tree will not deselect nodes.
      */
     disablePoppingOnBackspace?: boolean
+    /** dsiableKeyboardNavigation will disable keyboard navigation of the tree */
+    dsiableKeyboardNavigation?: boolean
```
mrchief, line 114:
```suggestion
    /** disableKeyboardNavigation will disable keyboard navigation of the tree */
    disableKeyboardNavigation?: boolean
```

src/index.js
```diff
@@ -41,7 +42,9 @@ class DropdownTreeSelect extends Component {
     onNodeToggle: PropTypes.func,
     onFocus: PropTypes.func,
     onBlur: PropTypes.func,
+    onSearchChange: PropTypes.func,
     mode: PropTypes.oneOf(['multiSelect', 'simpleSelect', 'radioSelect', 'hierarchical']),
+    pageSize: PropTypes.number,
```
mrchief, line 47: @gandhis1 I did think about it during the infinite scroll implementation but ultimately decided against it. Reasons being
 - an arbitrary pageSize doesn't help with the overall intent of optimization - e.g, setting a lower number is actually counter productive and setting a higher number will slow down the rendering. This gets exacerbated if you have multiple dropdowns on the page. Furthermore, unless you've a really big screen, you can't see more than 100 anyway so why hold it in DOM?

src/index.js
```diff
@@ -50,18 +53,21 @@ class DropdownTreeSelect extends Component {
     inlineSearchInput: PropTypes.bool,
     tabIndex: PropTypes.number,
     disablePoppingOnBackspace: PropTypes.bool,
+    disableKeyboardNavigation: PropTypes.bool,
```
mrchief, line 56: What's the use case for this? At a glance it seems like an a11y blocker but people don't have to use the keyboard if they don't want to so curios as to why this is needed.

gandhis1: From my PR description above:

> The new prop disableKeyboardNavigation allows disabling all actions associated with a key down event in the search input box. This is important if you need an input box to behave normally. For instance, when I press SHIFT+HOME, it should select the entire line, so I can press BACKSPACE to clear it.

I can't use arrow keys either on the search box right now.

mrchief: > an input box to behave normally

I hear you but it's not really a regular input box - hence the special treatment. 

> when I press SHIFT+HOME, it should select the entire line, so I can press BACKSPACE to clear it.

Can this handled by adding additional functionality to the keyboard navigation logic? So that we don't have to choose between having it or not at all?

charleslee94:
Hello! Curious on the status of this pr, this change would help out my use case of not clearing the search term on selection.

Thanks!

PR #604, opened 2022-11-23
Update react-dropdown-tree-select.d.ts

tsuzumiC:
Shouldn't children of TreeNodeProps also be of TreeNodeProps so that you can go deeper than one level?


gandhis1:
Changing the typescript definition won't do anything. You would need to also update the actual code.

mrchief:
This change looks ok to me but then I'm not sure why @gandhis1 thinks we need to update code. Typescript definitions are just syntactic sugar, they have no bearing on code.

mrchief:
I'm not suer if we had a resolution on this? Let me know if this needs to be reopened.

PR #641, opened 2023-07-19
feat: Call on input cleared when all characters are removed

### LINKED ISSUE #637: The searchTerm in the searchPredicate function is never an empty string
**Describe the bug**
When a user types in the input field, the searchPredicate function is triggered with whatever the user is writing. However, when the user empties his input field, the searchPredicate function is never triggered with an empty string.

**To Reproduce**
```
function searchPredicate(node, searchTerm) {
  console.log(searchTerm) // this is never an empty string ""
  return false
}

return <DropdownTreeSelect data={data} searchPredicate={searchPredicate} />
```

**Expected behavior**
When the input field is emptied, the searchPredicate function is triggered with an empty string.

lennertVanSever:
## What does it do?

Calls a new prop function 'onInputCleared" when all the characters are removed from the input box. Allows users to perform an action when the search box is cleared in an easy way.

## Fixes # (issue)

Closes #637 

## Type of change

New feature (non-breaking change which adds functionality)

# Checklist:

- [ ] I have commented my code, particularly in hard-to-understand areas => not applicable


lennertVanSever:
@mrchief can this pull request be brought back to life?

PR #644, opened 2023-08-23
Fix: Toggle node by clicking on label.

sanjayconnamara:
Fixing issue where node is expanding while clicking on label.


mrchief:
This is preventing standard behavior on HTML labels. What reasons are there for doing this?

